PR #52, opened 2016-05-05
encode sessionDataKey

mefarazath:
fixing https://wso2.org/jira/browse/IDENTITY-4552


mefarazath:
It's wrong to encode the sessionDataKey here since we have already added the sessionDataKey to cache in the line before. Hence closing this pull request

PR #112, opened 2016-07-15
Fixing IDENTITY-4811

DMHP:


DMHP:
Moved the change to registry by using https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/128

PR #138, opened 2016-08-08
Fixing Identity_4814

DMHP:
Removing "schema" parameter since [1] does not specify such parameter to the user info request.

[1] http://openid.net/specs/openid-connect-core-1_0.html#UserInfo


DMHP:
This is already fixed by  https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/143

PR #139, opened 2016-08-09
Fixing Identity 4840

DMHP:
Setting auth_time claim in Id Token


DMHP:
Fixed with : https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/147

PR #140, opened 2016-08-09
Fixing Identity-4823

DMHP:


DMHP:
Fixed with : https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/148

PR #150, opened 2016-08-15
Fix IDENTITY-4914 [master]

mefarazath:
* Adding 'openid' scope validation in IDTokenResponseValidator class.
* Extend TokenValidator in IDTokenResponseValidator.
* Remove duplicated code in IDTokenTokenResponseValidator by extending the IDTokenResponseValidator.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/IDTokenResponseValidator.java
+++ b/components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/IDTokenResponseValidator.java
@@ -19,19 +19,16 @@
 package org.wso2.carbon.identity.oauth.common;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.oltu.oauth2.as.validator.TokenValidator;
 import org.apache.oltu.oauth2.common.OAuth;
 import org.apache.oltu.oauth2.common.error.OAuthError;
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/IDTokenResponseValidator.java
```diff
@@ -58,4 +63,23 @@ public void validateMethod(HttpServletRequest request) throws OAuthProblemExcept
...
 
-}
\ No newline at end of file
+    /**
+     * Method to check whether the scope parameter string contains 'openid' as a scope.
+     *
+     * @param scope
+     * @return
+     */
+    private static boolean containOIDCScope(String scope) {
```
johannnallathamby: Do we have a util package in oauth.common bundle? If so can we move this there as a public static method? We have similar one in OAuth2Util.isOIDCAuthzRequest(). However since this is in oauth bundle we can't take a dependency on this.

mefarazath: We don't have a Util package in oauth.common bundle, hence I added this private method since it will only be used by this class.
 
Yes, the idea behind adding this util method in this was to avoid taking the dependency from oauth bundle.

PR #213, opened 2016-09-02
fixing IDENTITY-4962

GayanM:


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/authz/handlers/TokenResponseTypeHandler.java
```diff
@@ -408,7 +411,7 @@ private void buildIdToken(OAuthAuthzReqMessageContext msgCtx, OAuth2AuthorizeRes
             throws IdentityOAuth2Exception {
 
         if (StringUtils.isNotBlank(authzRespDTO.getAccessToken())) {
-            addUserAttributesToCache(authzRespDTO.getAccessToken(), msgCtx.getAuthorizationReqDTO());
+            addUserAttributesToCache(authzRespDTO.getAccessToken(), msgCtx.getAuthorizationReqDTO(), msgCtx);
```
Pushpalanka: We can just send the msgCtx as the second parameter, without reading out AuthorizationReqDTO from it.

PR #248, opened 2016-10-27
Invalid Refresh Token Error Handling

brucecloud:
We are using wso2 identity server 5.1.0 and came across the following situation:

-- client possesses refresh token generated with access token
-- access token is now beyond validity period, but is still in 'ACTIVE' state in database
-- refresh token is now beyond validity period as well
-- client posts refresh token and rather than getting a 400 bad request, a 500 internal server error is returned

Our testing shows that this is because the refresh grant handler is returning a response token data transfer object with error information and a null access token.  The access token issuer is not checking whether the response is in error and instead proceeds as if the access token was assigned.

The addition of the few lines of code in my commits below cause the proper 400 bad request response to be sent back to the client, and also update the refresh grant handler to log consistent with the access token issuer.

Apologies if I have compared to the wrong branch, etc, as I am not sure where you are in your release cycle or how you would prefer to receive these types of suggestions.


mefarazath:
Good catch @brucecloud !!!
Thanks for reporting the issue and sending the fix along with it :+1: . 
Will verify the fix and merge it with our next release of this component.

mefarazath:
Created https://wso2.org/jira/browse/IDENTITY-5296 to track this.

mefarazath:
@johannnallathamby can we merge this fix for 5.3.0? Will this require a major version bump?

darshanasbg:
Thanks @brucecloud for the contribution..!

PR #260, opened 2016-11-26
IDENTITY-5332 :  If User log out and log in again with the same user, IS will throw exception

godwinamila:
If User log out and log in again with the same user, IS will throw exception


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/SAMLAssertionClaimsCallback.java
```diff
@@ -184,7 +184,8 @@ public void handleCustomClaims(JWTClaimsSet jwtClaimsSet, OAuthAuthzReqMessageCo
                 getUserAttributesFromCache(requestMsgCtx.getProperty(OAuthConstants.ACCESS_TOKEN).toString());
         Map<String, Object> claims = Collections.emptyMap();
 
-        if (userAttributes.isEmpty() && requestMsgCtx.getProperty(OAuthConstants.AUTHZ_CODE) != null) {
+        if ((userAttributes == null || userAttributes.isEmpty())
```
mefarazath: Shall we use `MapUtils.isEmpty(userAttributes)` to be consistent with the other places in the same class?

PR #282, opened 2016-12-13
IDENTITY-5476: UserInfo Error Response MUST include WWW-Authenticate header

madurangasiriwardena:


madurangasiriwardena:
When the application is revoked, tokens of the application are also revoked. So application active state need not to be checked in OpenIDConnectUserEndpoint. This will reduce 2 unnecessary database calls.

PR #288, opened 2016-12-14
Fixed IDENTITY-5473

IndunilRathnayake:


IndunilRathnayake:
The user store domain and tenant domain has to be appended to subject in the framework.

PR #312, opened 2017-01-03
Get username from authenticated user instead of carbon context in DCR…

IsuraD:
… owner


IsuraD:
close this due to cyclic dependancy

PR #328, opened 2017-04-17
IDENTITY-5870: Handle expired token correctly in UserInfo flow

mefarazath:
* Handle inactive/expired token when retrieve claims in userinfo.
* Improve token retrieval logic in ClaimUtil class.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/util/ClaimUtil.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/util/ClaimUtil.java
@@ -23,6 +23,7 @@ import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.oltu.oauth2.common.error.OAuthError;
 import org.wso2.carbon.identity.application.authentication.framework.exception.FrameworkException;
 import org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants;
 import org.wso2.carbon.identity.application.common.model.ClaimMapping;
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/util/ClaimUtil.java
```diff
@@ -83,13 +83,16 @@ private ClaimUtil() {
...
+                    .getAuthorizationContextToken().getTokenString());
             ApplicationManagementService applicationMgtService = OAuth2ServiceComponentHolder.getApplicationMgtService();
-            String clientId = null;
+            String clientId;
             if (accessTokenDO != null) {
                 clientId = accessTokenDO.getConsumerKey();
+            } else {
+                // this means the token is not active so we can't proceed further
+                throw new UserInfoEndpointException(OAuthError.ResourceResponse.INVALID_TOKEN,
+                        "Invalid Access Token. Access token is not ACTIVE.");
```
ruwanta: @Pushpalanka It is better to say what is the exact error in the exception. e.g.
"Could not find the AccessToken corresponding to the token string xyz. The token may be invalid or not ACTIVE"

mefarazath: @ruwanta at this point the only way this exception could occur is if the token is in the database but not in the ACTIVE state. If there was no AccessToken corresponding to a particular string xyz in the DB we wouldn't have come up to this point. ie. it would have failed at this [check](https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java#L70).

ruwanta: @mefarazath @Pushpalanka The explanation is sufficient.

PR #331, opened 2017-04-18
IDENTITY-4976: Handle TimeSkew properly

mefarazath:
Found few more places where the timeSkew calculation is incorrect.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
@@ -446,19 +446,20 @@ public class OAuth2Util {
 
         long validityPeriodMillis = accessTokenDO.getValidityPeriodInMillis();
         long issuedTime = accessTokenDO.getIssuedTime().getTime();
-        long currentTime = System.currentTimeMillis();
 
         //check the validity of cached OAuth2AccessToken Response
-        long skew = OAuthServerConfiguration.getInstance().getTimeStampSkewInSeconds() * 1000;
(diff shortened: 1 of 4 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -450,13 +450,13 @@ public static AccessTokenDO validateAccessTokenDO(AccessTokenDO accessTokenDO) {
...
+        if (issuedTime + validityPeriodMillis - (currentTime - skew) > 1000) {
             long refreshValidity = OAuthServerConfiguration.getInstance()
                     .getRefreshTokenValidityPeriodInSeconds() * 1000;
-            if (issuedTime + refreshValidity - currentTime + skew > 1000) {
+            if (issuedTime + refreshValidity - currentTime - skew > 1000) {
                 //Set new validity period to response object
-                accessTokenDO.setValidityPeriod((issuedTime + validityPeriodMillis - (currentTime + skew)) / 1000);
-                accessTokenDO.setValidityPeriodInMillis(issuedTime + validityPeriodMillis - (currentTime + skew));
+                accessTokenDO.setValidityPeriod((issuedTime + validityPeriodMillis - (currentTime - skew)) / 1000);
+                accessTokenDO.setValidityPeriodInMillis(issuedTime + validityPeriodMillis - (currentTime - skew));
```
ruwanta: (issuedTime + validityPeriodMillis - (currentTime - skew))
Need to calculate this value at only one place, and assign a local variable such as "validForMillis" and use that variable in lines 453, 458, 459.
Having the same code duplicated in multiple places is a not correct practice.

PR #336, opened 2017-04-24
IDENTITY-5830 : move cache key string creation logic

denuwanthi:


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
```diff
@@ -103,13 +103,13 @@ public boolean validateGrant(OAuthTokenReqMessageContext tokReqMsgCtx) throws Id
             String scope = OAuth2Util.buildScopeString(authzCodeDO.getScope());
             String authorizedUser = authzCodeDO.getAuthorizedUser().toString();
             boolean isUsernameCaseSensitive = IdentityUtil.isUserStoreInUsernameCaseSensitive(authorizedUser);
-            String cacheKeyString;
-            if (isUsernameCaseSensitive) {
-                cacheKeyString = clientId + ":" + authorizedUser + ":" + scope;
-            } else {
-                cacheKeyString = clientId + ":" + authorizedUser.toLowerCase() + ":" + scope;
-            }
             if (cacheEnabled) {
```
madurangasiriwardena: You can move the cacheEnabled condition also to the outer if statement.

denuwanthi: @madurangasiriwardena 
Seems like if we do that we are not going to get the invalid_grant error when resend the authorization code during cache disabled scenario.
In [1] in check for 'isValidGrant'. This is returned by this method validateGrant.
In cache disable case, if we put the cacheEnabled variable to the upper if clause, it will never go inside the if statement and will not return false.
So, the token is not revoked.


[1]https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java#L220

PR #348, opened 2017-05-08
Fix  Identity-4915:'sub' claim not returned in id_tokens generated with Client credential grant type

DMHP:
Remove ID Token issuing and claims returning from User info endpoint in client credential grant type.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java
@@ -95,10 +95,6 @@ public class OpenIDConnectUserEndpoint {
         } catch (OAuthSystemException e) {
             log.error("UserInfoEndpoint Failed", e);
             throw new OAuthSystemException("UserInfoEndpoint Failed");
-        } catch (IdentityOAuthAdminException e) {
-            if (log.isDebugEnabled()) {
-                log.debug("Error in getting grant type.", e);
-            }
         }
 
         ResponseBuilder respBuilder =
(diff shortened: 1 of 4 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java
```diff
@@ -76,18 +79,26 @@ public Response getUserClaims(@Context HttpServletRequest request) throws OAuthS
...
+            if (!GrantType.CLIENT_CREDENTIALS.toString().equals(grantType)) {
+                UserInfoResponseBuilder userInfoResponseBuilder =
+                        UserInfoEndpointConfig.getInstance().getUserInfoResponseBuilder();
+                response = userInfoResponseBuilder.getResponseString(tokenResponse);
+            }
 
         } catch (UserInfoEndpointException e) {
             return handleError(e);
         } catch (OAuthSystemException e) {
             log.error("UserInfoEndpoint Failed", e);
```
sagara-gunathunga: @DMHP What is the exact logic here ? I mean why are we logging original exception and return a new type of a exception with lesser details ?

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java
```diff
@@ -76,18 +79,26 @@ public Response getUserClaims(@Context HttpServletRequest request) throws OAuthS
...
             throw new OAuthSystemException("UserInfoEndpoint Failed");
+        } catch (IdentityOAuthAdminException e) {
+            if (log.isDebugEnabled()) {
+                log.debug("Error in getting grant type.", e);
+            }
```
sagara-gunathunga: @DMHP If debug is not enabled we don't get any error details, why ? 

Additionally within the same try-catch block we have used 3 different styles, we have to use one consistent style in all the places.

DMHP: @sagara-gunathunga Mistakenly I have put a debug log here. I have already fixed it. Anyway I feel throwing server side exceptions from this endpoint is incorrect. According to the spec we should  generate a proper json error and return it.

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java
DMHP: @sagara-gunathunga  I feel throwing server side exceptions from this endpoint is incorrect. According to the spec we should  generate a proper json error and return it. As this is an existing code and we have done it in several places I will create a seperate jira to handle this.

DMHP:
Changed the logic and fixed from https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/358

PR #381, opened 2017-06-01
IDENTITY-6032: Pass HTTP Request Headers to Grant Handlers

mefarazath:
https://wso2.org/jira/browse/IDENTITY-6032
(This involves API additions)

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/model/HttpRequestHeader.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/model/HttpRequestHeader.java
@@ -1,0 +1,44 @@
+package org.wso2.carbon.identity.oauth2.model;
+
+import java.io.Serializable;
+
+/**
+ * This class is used to store http request header information
+ */
+public class HttpRequestHeader implements Serializable {
+
+    private static final long serialVersionUID = 5419655486789962879L;
+
+    private String name;
+    private String[] values;
+
+    /**
+     * Instantiate a HTTPHeader object for the given name and values
+     *
+     * @param name    header name
+     * @param values parameter values
+     */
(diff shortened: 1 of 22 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/model/HttpRequestHeader.java
```diff
@@ -0,0 +1,42 @@
+package org.wso2.carbon.identity.oauth2.model;
+
+import java.io.Serializable;
+
+/**
+ * This class is used to store http request header information
+ */
+public class HttpRequestHeader implements Serializable {
```
Pushpalanka: Serializable class should have a static final serialVersionUID field of type long. In IDEA there is a plugin to auto generate this.

mefarazath:
Added **serialVersionUID** to the HttpRequestHeader class.

mefarazath:
@isharak @Pushpalanka @kasunbg 

Today I during a discussion related to implementing token binding I realized that we need to pass headers to underlying grant handlers and other extensions points. 

According to what I understood in TLS binding, we make use of information sent in headers to validate refresh token etc. So can we consider merging this PR to facilitate passing header information to downstream extensions from the TokenEndpoint?

kasunbg:
Hi @mefarazath,

Yes, we are implementing TLS Token Binding with a [TTRP](https://tools.ietf.org/html/draft-ietf-tokbind-ttrp-01) in the first phase. In this, a HTTP header named "Sec-Provided-Token-Binding-Id" and "Sec-Referred-Token-Binding-Id" will be sent to the back-end Identity Server. 

These headers need to be validated by the grant types. So, the grant type handlers must receive these HTTP headers.

Thanks.

kasunbg:
This is an API addition, and the existing signatures are not changed. So, we should do a minor version bump when merging this PR.

PR #368, opened 2017-05-24
Fix IDENTITY-5959 :Incorrect error message for OAuth Token Request with a non e…

DMHP:
…xisitng client_id

PR #382, opened 2017-06-01
IDENTITY-6039: Fix incorrect error message for OAuth Authz Request for a non-exisiting client_id

mefarazath:
https://wso2.org/jira/browse/IDENTITY-6039


mefarazath:
This is similar to https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/368 done for the token endpoint.

PR #403, opened 2017-07-05
IDENTITY-6088: fix for returning invalid id token without required cl…

denuwanthi:
…aims

public jira : https://wso2.org/jira/browse/IDENTITY-6088

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
  *
  * WSO2 Inc. licenses this file to you under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except
@@ -18,6 +18,9 @@
 
 package org.wso2.carbon.identity.oauth2.token;
 
+import com.nimbusds.jwt.JWT;
+import com.nimbusds.jwt.JWTParser;
+import com.nimbusds.jwt.ReadOnlyJWTClaimsSet;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.oltu.oauth2.common.error.OAuthError;
@@ -29,6 +32,7 @@ import org.wso2.carbon.identity.oauth.cache.AppInfoCache;
 import org.wso2.carbon.identity.oauth.cache.AuthorizationGrantCache;
 import org.wso2.carbon.identity.oauth.cache.AuthorizationGrantCacheEntry;
 import org.wso2.carbon.identity.oauth.cache.AuthorizationGrantCacheKey;
+import org.wso2.carbon.identity.oauth.common.OAuth2ErrorCodes;
 import org.wso2.carbon.identity.oauth.common.OAuthConstants;
 import org.wso2.carbon.identity.oauth.common.exception.InvalidOAuthClientException;
 import org.wso2.carbon.identity.oauth.config.OAuthServerConfiguration;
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -428,4 +447,34 @@ private void setResponseHeaders(OAuthTokenReqMessageContext tokReqMsgCtx,
...
         }
     }
+
+    /**
+     * Method to check whether id token contains the required claims(iss,sub,aud,exp,iat) defined by the oidc spec
+     * @param jwt id token
+     * @return true or false(whether id token contains the required claims)
+     * @throws ParseException
+     */
+    private boolean isValidIdToken(JWT jwt) throws ParseException {
```
kasunbg: Since this method is responsible for actually validating the token, we should not have a ParseException in the method signature. The return boolean status can be used instead.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -305,7 +310,21 @@ public OAuth2AccessTokenRespDTO issue(OAuth2AccessTokenReqDTO tokenReqDTO)
 
         if (tokReqMsgCtx.getScope() != null && OAuth2Util.isOIDCAuthzRequest(tokReqMsgCtx.getScope())) {
             IDTokenBuilder builder = OAuthServerConfiguration.getInstance().getOpenIDConnectIDTokenBuilder();
-            tokenRespDTO.setIDToken(builder.buildIDToken(tokReqMsgCtx, tokenRespDTO));
+            String idToken = builder.buildIDToken(tokReqMsgCtx, tokenRespDTO);
+            try {
+                JWT jwt = JWTParser.parse(idToken);
```
kasunbg: The DefaultIDTokenBuilder#buildIDToken method actually creates the JWT token object, and then serialize it. [1] In the above code, we take that serialized string and parse it again into a JWT token object. 

IMO, this is not necessary and will introduce additional latency.

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/6855d9a8bd26c74b2cdd6bdf02ccd2be7ba75eea/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultIDTokenBuilder.java#L316

denuwanthi: @kasunbg 
If the token is invalid here I return tokenRespDTO = handleError(
                        OAuth2ErrorCodes.SERVER_ERROR,
                        "Server Error",
                        request.getOauth2AccessTokenReqDTO());
Currently in DefaultIDTokenBuilder, it returns PlainJWT[1] or signJWT[2]. So, If I do the validation for the token there, I need to return some String to indicate the invalid token.
Shall I send some error string or 'tokenRespDTO.getErrorMsg()' ?

Appreciate some suggestion on this.

[1]https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/6855d9a8bd26c74b2cdd6bdf02ccd2be7ba75eea/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultIDTokenBuilder.java#L316
[2]https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/6855d9a8bd26c74b2cdd6bdf02ccd2be7ba75eea/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultIDTokenBuilder.java#L318

kasunbg: How about using the exception handling capabilities for this.
This method has a throws clause for IdentityOAuth2Exception.
 
So, how abt creating a sub-class of IdentityOAuth2Exception say, IDTokenValidationFailureException? It gets thrown when the validation fails. 
We can catch and handle this.

We need to special-case the default ID Token builder to overcome possible penalty on performance of building the id token twice.

denuwanthi:
Updated PR at https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/427

PR #407, opened 2017-07-11
IDENTITY-6133: Add the self-signed JWT access token issuer.

JKAUSHALYA:
JIRA: https://wso2.org/jira/browse/IDENTITY-6133

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
@@ -33,13 +33,13 @@ import org.apache.oltu.oauth2.common.exception.OAuthSystemException;
 import org.wso2.carbon.base.MultitenantConstants;
 import org.wso2.carbon.core.util.KeyStoreManager;
 import org.wso2.carbon.identity.application.authentication.framework.model.AuthenticatedUser;
-import org.wso2.carbon.identity.application.common.model.ClaimMapping;
 import org.wso2.carbon.identity.core.util.IdentityTenantUtil;
 import org.wso2.carbon.identity.oauth.config.OAuthServerConfiguration;
 import org.wso2.carbon.identity.oauth2.IdentityOAuth2Exception;
 import org.wso2.carbon.identity.oauth2.authz.OAuthAuthzReqMessageContext;
 import org.wso2.carbon.identity.oauth2.config.SpOAuth2ExpiryTimeConfiguration;
 import org.wso2.carbon.identity.oauth2.util.OAuth2Util;
+import org.wso2.carbon.identity.openidconnect.CustomClaimsCallbackHandler;
 
 import java.security.Key;
 import java.security.interfaces.RSAPrivateKey;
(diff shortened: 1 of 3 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
```diff
@@ -0,0 +1,370 @@
...
+     * @throws IdentityOAuth2Exception
+     */
+    protected String buildJWTToken(OAuthTokenReqMessageContext request)
+            throws IdentityOAuth2Exception {
+
+        // Set claims to jwt token.
+        JWTClaimsSet jwtClaimsSet = createJWTClaimSet(null, request, request.getOauth2AccessTokenReqDTO()
+                .getClientId());
+
+        if (Arrays.asList((request.getScope())).contains(AUDIENCE)) {
```
mefarazath: How is scope related to 'aud' value?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
```diff
@@ -0,0 +1,370 @@
...
+                tenantDomain = tokenContext.getAuthorizedUser().getTenantDomain();
+            } else if (authorizationContext != null) {
+                tenantDomain = authorizationContext.getAuthorizationReqDTO().getUser().getTenantDomain();
+            }
+
+            if (tenantDomain == null) {
+                throw new IdentityOAuth2Exception("Cannot resolve the tenant domain of the user.");
+            }
+
+            int tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
```
mefarazath: What if someone wants all the access token to be signed with private key of tenant of the App rather than the user?
We have a property to decide this for id_tokens "SignJWTWithSPKey" [1]


[1] https://wso2.org/jira/browse/IDENTITY-6036

JKAUSHALYA: Is this implemented ? Do we have the facility retrieve the certificate/private key of the SP ?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
JKAUSHALYA: Seems like we are getting audiences in the scope array.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
mefarazath: No what i mean was for a SaaS app the user's tenant and SP's tenant might be different. Say we create an app in carbon.super and user comes from foo and bar tenants. Using this property we always sign with carbon.super tenant key regardless of the tenant of the user

PR #420, opened 2017-07-28
IDENTITY-6155: Fix invoking the user info endpoint without properly setting the 'Bearer' header causes server errors

hasinidilanka:


hasinidilanka:
@isharak @omindu @mefarazath @darshanasbg

PR #428, opened 2017-08-09
IDENTITY-6175: Fixing perf penalty in auth code revocation logic

kasunbg:

IDENTITY-6175: Fixing perf penalty that happened to the iterated revocation of auth codes when re-generating client secrets

Immediate

### Follow up actions

N/A

#### General

- [x] **Is this PR explained thoroughly?** All code changes must be accounted for in the PR description.
- [x] **Is the PR labeled correctly?**

#### Functionality

- [x] **Are all requirements met?** Compare implemented functionality with the requirements specification.
- [x] **Does the UI work as expected?** There should be no Javascript errors in the console; all resources should load. There should be no unexpected errors. Deliberately try to break the feature to find out if there are corner cases that are not handled.

#### Code

- [x] **Do you fully understand the introduced changes to the code?** If not ask for clarification, it might uncover ways to solve a problem in a more elegant and efficient way.
- [x] **Does the PR introduce any inefficient database requests?** Use the debug server to check for duplicate requests.
- [x] **Are all necessary strings marked for translation?** All strings that are exposed to users via the UI must be [marked for translation](https://docs.djangoproject.com/en/1.10/topics/i18n/translation/).

#### Tests

- [x] **Are there sufficient test cases?** Ensure that all components are tested individually; models, forms, and serializers should be tested in isolation even if a test for a view covers these components.
- [x] **If this is a bug fix, are tests for the issue in place?**  There must be a test case for the bug to ensure the issue won’t regress. Make sure that the tests break without the new code to fix the issue.
- [x] **If this is a new feature or a significant change to an existing feature?** has the manual testing spreadsheet been updated with instructions for manual testing?

#### Security

- [x] **Confirm this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets.**
- [x] **Are all UI and API inputs run through forms or serializers?**
- [x] **Are all external inputs validated and sanitized appropriately?**
- [x] **Does all branching logic have a default case?**
- [x] **Does this solution handle outliers and edge cases gracefully?**
- [x] **Are all external communications secured and restricted to SSL?**

#### Documentation

- [x] **Are changes to the UI documented in the platform docs?** If this PR introduces new platform site functionality or changes existing ones, the changes should be documented.
- [x] **Are changes to the API documented in the API docs?** If this PR introduces new API functionality or changes existing ones, the changes must be documented.
- [x] **Are reusable components documented?** If this PR introduces components that are relevant to other developers (for instance a mixin for a view or a generic form) they should be documented in the Wiki.


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/SQLQueries.java
```diff
@@ -197,6 +197,10 @@
             "IDN_OAUTH2_AUTHORIZATION_CODE WHERE CONSUMER_KEY_ID IN (SELECT ID FROM IDN_OAUTH_CONSUMER_APPS WHERE " +
             "CONSUMER_KEY = ?) AND STATE = ?";
 
+    public static final String UPDATE_AUTHORIZATION_CODE_STATE_FOR_CONSUMER_KEY = "UPDATE "
+            + "IDN_OAUTH2_AUTHORIZATION_CODE SET STATE=? WHERE CONSUMER_KEY_ID IN (SELECT ID FROM "
```
madurangasiriwardena: Will there be a performance improvement if we add STATE='ACTIVE' also to the where clause?

kasunbg: parameter substitution happens when constructing the query. So, the comparatively the gain is low. Further more, this method is useful later when we want to bulk update the AUTH codes.

madurangasiriwardena: I was thinking about changing the query like below.
UPDATE IDN_OAUTH2_AUTHORIZATION_CODE SET STATE=? WHERE CONSUMER_KEY_ID IN (SELECT ID FROM IDN_OAUTH_CONSUMER_APPS WHERE CONSUMER_KEY = ?) AND STATE='ACTIVE'
Will it gain any performance improvement?

kasunbg: You added a new condition here right? DB will not update a row if the current and the new value are the same. 
I tested the latency values empirically in MySQL. Both queries yield almost the same latency. So, I would say the perf gain/loss is negligible.

PR #602, opened 2017-10-22
Refactor OAuthAppDAO test cases and add unit tests for JWTTokenIssuer

mefarazath:
- Refactor OAuthAppDAO test class.
- Remove redundant test util class.
- Add unit tests for JWTTokenIssuer.
- Minor code improvements on unit test cases.

### Follow up actions
- Trigger a sonar build


components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth/dao/OAuthAppDAOTest.java
```diff
@@ -579,13 +579,13 @@ public void testGetAppInformationByAppName(Boolean enablePKCE) throws Exception
         mockStatic(OAuth2ServiceComponentHolder.class);
         when(OAuth2ServiceComponentHolder.isPkceEnabled()).thenReturn(enablePKCE);
 
-        TestUtils.initPrivilegedCarbonContext("PRIMARY", -1234, USER_NAME);
+        CommonTestUtils.initPrivilegedCarbonContext("PRIMARY", -1234, USER_NAME);
```
ashensw: Better to use `carbon.super` instead of `PRIMARY` since the parameter is `tenantDomain`. Or we can define another method in the CommonTestUtils with a different signature where we have only the `username`?

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth/dao/OAuthAppDAOTest.java
```diff
@@ -605,13 +605,13 @@ public void testGetAppInformationByAppNameWithExceptions(Boolean enablePKCE) thr
...
```
ashensw: Better to use `carbon.super` instead of `PRIMARY` since the parameter is `tenantDomain`. Or we can define another method in the CommonTestUtils with a different signature where we have only the `username`?

PR #662, opened 2017-11-09
Refactored the OAuth2ScopeService without powermock

supunharsha:

Refactord the Test class **OAuth2ScopeServiceTest**

Immediately

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/cache/OAuthScopeCache.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/cache/OAuthScopeCache.java
@@ -73,7 +73,7 @@ public class OAuthScopeCache extends BaseCache<OAuthScopeCacheKey, Scope> {
         if (IdentityUtil.getIdentityCacheConfig(IDENTITY_CACHE_MANAGER, OAUTH_SCOPE_CACHE_NAME).isEnabled()) {
             entry = super.getValueFromCache(key);
             if (log.isDebugEnabled()) {
-                log.debug("Scope is getting from the cache. \n" + entry.toString());
+                log.debug("Scope is getting from the cache. \n" + String.valueOf(entry));
             }
         }
         return entry;
--- a/components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/OAuth2ScopeServiceTest.java
+++ b/components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/OAuth2ScopeServiceTest.java
@@ -16,57 +16,34 @@
 
 package org.wso2.carbon.identity.oauth2;
 
-import org.mockito.InjectMocks;
-import org.mockito.Mock;
-import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.apache.commons.lang.StringUtils;
+import org.powermock.reflect.Whitebox;
+import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 import org.wso2.carbon.identity.base.IdentityException;
-import org.wso2.carbon.identity.core.model.IdentityCacheConfig;
-import org.wso2.carbon.identity.core.util.IdentityDatabaseUtil;
 import org.wso2.carbon.identity.core.util.IdentityUtil;
 import org.wso2.carbon.identity.oauth.cache.OAuthScopeCache;
 import org.wso2.carbon.identity.oauth.cache.OAuthScopeCacheKey;
 import org.wso2.carbon.identity.oauth2.bean.Scope;
-import org.wso2.carbon.identity.oauth2.dao.ScopeMgtDAO;
-import org.wso2.carbon.identity.oauth2.test.utils.CommonTestUtils;
+import org.wso2.carbon.identity.oauth2.util.Oauth2ScopeUtils;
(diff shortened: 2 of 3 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/cache/OAuthScopeCache.java
```diff
@@ -72,7 +72,7 @@ public Scope getValueFromCache(OAuthScopeCacheKey key) {
         Scope entry = null;
         if (IdentityUtil.getIdentityCacheConfig(IDENTITY_CACHE_MANAGER, OAUTH_SCOPE_CACHE_NAME).isEnabled()) {
             entry = super.getValueFromCache(key);
-            if (log.isDebugEnabled()) {
+            if (log.isDebugEnabled() && entry != null) {
                 log.debug("Scope is getting from the cache. \n" + entry.toString());
```
dharshanaw: Use String.valueOf() instaid of null check.

supunharsha: done

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/OAuth2ScopeServiceTest.java
```diff
@@ -79,260 +56,95 @@
...
 
     @Test(expectedExceptions = IdentityException.class)
     public void testRegisterScopeWithNoScopeName() throws Exception {
         String name = "";
         String description = "dummyScopeDescription";
-        mockedScope = new Scope(name, description);
-
-        oAuth2ScopeService.registerScope(mockedScope);
+        Scope scope = new Scope(name, description);
+        oAuth2ScopeService.registerScope(scope);
```
danushkaf: Shall we assert for a value after calling get scope. Else it would only break if it throw an exception. Applicable for all test methods

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/OAuth2ScopeServiceTest.java
```diff
@@ -79,260 +56,95 @@
...
-        doThrow(new IdentityOAuth2ScopeServerException("dummyException")).when(mockedScopeMgtDAO).updateScopeByName
-                (any(Scope.class), anyInt());
-
-        oAuth2ScopeService.updateScope(mockedScope);
+    @Test(expectedExceptions = IdentityOAuth2ScopeException.class)
+    public void testDeleteScope() throws Exception {
+        Scope dummyScope = new Scope(SCOPE_NAME, SCOPE_DESCRIPTION);
+        oAuth2ScopeService.registerScope(dummyScope);
+        oAuth2ScopeService.deleteScope(SCOPE_NAME);
+        oAuth2ScopeService.getScope(SCOPE_NAME);
```
danushkaf: We need to assert here for the value we get.

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/OAuth2ScopeServiceTest.java
```diff
@@ -79,260 +56,95 @@
...
+    public void testGetScope(boolean existWithinCache) throws Exception {
+        Scope dummyScope = new Scope(SCOPE_NAME, SCOPE_DESCRIPTION);
+        oAuth2ScopeService.registerScope(dummyScope);
+        if (!existWithinCache) {
+            OAuthScopeCache.getInstance().clearCacheEntry(new OAuthScopeCacheKey(SCOPE_NAME, Integer.toString(
+                    Oauth2ScopeUtils.getTenantID())));
+        }
+        assertEquals(oAuth2ScopeService.getScope(SCOPE_NAME).getName(), SCOPE_NAME, "Retrieving registered scope is " +
+                "failed");
+        oAuth2ScopeService.deleteScope(SCOPE_NAME);
```
danushkaf: May be we can call delete in @AfterMethod

supunharsha: Hi Danushka, we can't do this in tear down method since it is not common for all method. For example when it execute a test written for deleteScope method, and at the tearDown if it tries to delete again it throws a NPE.

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/OAuth2ScopeServiceTest.java
supunharsha: If it throws an exception, no line get executed after line 148. So assertion would not be happened. If it doesn't throws an exception then it means a test failure. Even if we assert the output from getScope method, still we expect that test should throw an exception based on the input. So still test should fail regardless of output in assertion.

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/OAuth2ScopeServiceTest.java
supunharsha: Hi Danushka, if this method doesn't throw an exception, it is a test failure. Suppose if we try to assert whether scope is there using getScope method,  and if it is there or not, still we are expecting to throw the exception. That means it is still a test failure.

PR #671, opened 2017-11-11
Minor refactoring to OIDCClaimUtil

mefarazath:


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OIDCClaimUtil.java
```diff
@@ -168,7 +168,7 @@ private static String getOIDCClaimUri(String scopeClaim) {
                 StringUtils.substringBefore(scopeClaim, SCOPE_CLAIM_PREFIX);
     }
 
-    protected static boolean isNotEmpty(Properties properties) {
+    private static boolean isNotEmpty(Properties properties) {
```
ruwanta: Is it ok to make this private ? Can there be a class extended from this Util ?

mefarazath: Planning to move the logic in the Util class to an extension. So that anyone can extend and override the behaviour of the ClaimUtil class.

PR #687, opened 2017-11-16
Fix id_token expiry time format

mefarazath:
- Fixing id_token expiry time calculation logic.
- Change default fallback value for id_token expiry

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
@@ -162,10 +162,8 @@ public class OAuthServerConfiguration {
     private String openIDConnectIDTokenIssuerIdentifier = null;
     private String openIDConnectIDTokenSubClaim = "http://wso2.org/claims/fullname";
     private String openIDConnectSkipUserConsent = "true";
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -914,8 +912,8 @@ public String getOpenIDConnectIDTokenExpiration() {
      *
      * @return ID Token expiry time in milliseconds.
      */
-    public long getOpenIDConnectIDTokenExpiryTimeInMillis() {
-        return openIDConnectIDTokenExpiryTimeInMillis;
+    public long getOpenIDConnectIDTokenExpiryTimeInSeconds() {
```
nuwandiw: Already have one API (getOpenIDConnectIDTokenExpiration) to get the value in seconds. Can't we keep `getOpenIDConnectIDTokenExpiryTimeInMillis` and return `openIDConnectIDTokenExpiryTimeInSeconds * 1000L`

mefarazath: I was thinking the same. Since the other expiry times in seconds are returned as int wanted to maintain consistency...

getOpenIDConnectIDTokenExpiration returns the value as a String

PR #778, opened 2018-02-01
Fixed IDENTITY-7312

IndunilRathnayake:

1. Updated user attributes are not cached against access token

- In federated scenario, if the claims have been updated, those are properly reflected in AuthorizationGrantCache against the authorization code.
- But when caching the user attributes against the access token in AuthorizationGrantCache, if the access token is already exists, we are not considering the attributes saved against the authorization code.
- So if the previously issued access token has not invalidated from cache, updated user claims are not added into the cache.
- We need to consider the entry against authorization code when caching.

2. Updated user attributes are not reflected in ID token
- User attributes are added to the AuthorizationGrantCache against access token, after generating the ID token.
- When generating the ID token, we get user attributes from the AuthorizationGrantCache which are stored against access token.
- So if the previously issued access token is not invalidated in the cache, every time we are issuing ID token with invalid set of user claims where the updated claims are not reflected.
- We need to add the user attributes to AuthorizationGrantCache against access token, before generating the ID token.

can merge this pull request.]

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -324,6 +324,10 @@ public OAuth2AccessTokenRespDTO issue(OAuth2AccessTokenReqDTO tokenReqDTO)
                     tokReqMsgCtx.getAuthorizedUser() + " and scopes: " + tokenRespDTO.getAuthorizedScopes());
         }
 
+        if (tokenReqDTO.getGrantType().equals(GrantType.AUTHORIZATION_CODE.toString())) {
```
ruwanta: Cant we use 
if (grantType.equals(GrantType.AUTHORIZATION_CODE.toString())) {

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -421,17 +421,13 @@ private void addUserAttributesToCache(OAuth2AccessTokenReqDTO tokenReqDTO, OAuth
...
-                    if (log.isDebugEnabled()
-                            && IdentityUtil.isTokenLoggable(IdentityConstants.IdentityTokens.ACCESS_TOKEN)) {
-                        log.debug("No AuthorizationGrantCache entry found for the access token(hashed):" +
-                                DigestUtils.sha256Hex(newCacheKey.getUserAttributesId()) + ", hence adding to cache");
-                    }
-                    AuthorizationGrantCache.getInstance().addToCacheByToken(newCacheKey, authorizationGrantCacheEntry);
-                    AuthorizationGrantCache.getInstance().clearCacheEntryByCode(oldCacheKey);
-                } else {
-                    //if the user attributes are already saved for access token, no need to add again.
+                if (log.isDebugEnabled()
```
ruwanta: if (log.isDebugEnabled()){
 +                    if(IdentityUtil.isTokenLoggable(IdentityConstants.IdentityTokens.ACCESS_TOKEN)) {
 +                    log.debug("Adding AuthorizationGrantCache entry for the access token(hashed):" +
 +                            DigestUtils.sha256Hex(newCacheKey.getUserAttributesId()));
}else {
log.debug("Adding AuthorizationGrantCache entry for the access token(hashed):"
}
                  }

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
lahirus: +1,  there is another place to fix from old code. "log.debug("Client Id: " + tokenReqDTO.getClientId() + " is not authorized to use grant type: " +
                        tokenReqDTO.getGrantType());"

PR #782, opened 2018-02-05
Fixed the issue #2220

GDRDABARERA:
### Description

Self-contained access token OOTB broken since the DB column length is not enough. Therefore, when storing the self-contained access token it gives an exception.  Fixed this issue by hashing the  access token.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
@@ -25,6 +25,7 @@ import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.oltu.oauth2.common.exception.OAuthSystemException;
 import org.wso2.carbon.identity.application.authentication.framework.model.AuthenticatedUser;
 import org.wso2.carbon.identity.application.common.IdentityApplicationManagementException;
 import org.wso2.carbon.identity.application.common.model.ServiceProvider;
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
```diff
@@ -111,6 +112,19 @@ private void insertAccessToken(String accessToken, String consumerKey, AccessTok
...
         }
 
+        if (OAuthServerConfiguration.getInstance().usePersistedAccessTokenAlias()) {
+            try {
+                accessToken = OAuthServerConfiguration.getInstance().getIdentityOauthTokenIssuer()
+                        .getAccessTokenHash(accessToken);
+            } catch (OAuthSystemException e) {
+                if (log.isDebugEnabled() &&
+                        IdentityUtil.isTokenLoggable(IdentityConstants.IdentityTokens.ACCESS_TOKEN)) {
+                    log.debug("Error while getting access token hash for token: " + accessToken);
```
JKAUSHALYA: Even though we check isTokenLoggable, better if we log the access token as a hash. Not the direct value. Please check how we did it in other places.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
```diff
@@ -111,6 +112,20 @@ private void insertAccessToken(String accessToken, String consumerKey, AccessTok
                     "Authorized user should be available for further execution.");
         }
 
+        if (OAuthServerConfiguration.getInstance().usePersistedAccessTokenAlias()) {
+            try {
+                accessToken = OAuthServerConfiguration.getInstance().getIdentityOauthTokenIssuer()
```
GayanM: change variable name accessToken to some other name since you change the structure of original value

darshanasbg: @GayanM would you mind sending a PR for the mentioned improvement?

GDRDABARERA: I don't think it is really necessary to change the access token name because if the access token is self-signed only we get the hash as access token, and the hash is stored in the access token column. AFAIK, If we changed the name, we need to call insertAccessToken(accessToken, consumerKey, newAccessTokenDO, connection, userStoreDomain); method also twice.

PR #791, opened 2018-02-08
Integrate DB backed Service Provider public certs with Request Object Signature validation

mefarazath:
-  In our previous iterations the public certificate of the Service Provider (OAuth Client) used to validate signature of the Request Object was retrieved from the tenant keystore. (ie. It was imported to the tenant keystore with the alias set to client_id). 

Since we now have the capability to store the Service Provider public certificate in the DB (From improvements done in https://github.com/wso2/carbon-identity-framework/pull/1304) this PR improves the request object signature validation logic to retrieve the public cert from DB rather the key store.

This essentially means we are going to stop supporting public certs of Service Providers imported to keystores and will only considering public certs added to the DB. This will be a **backward incompatible API removal**.

In addition below improvements was also added,

- Add util methods to reuse retrieving public certs of Service Providers using client_id.
- Fix possible NPE due to error codes not set in exceptions.

### Improvements to be done
- Provide an option at Service Provider OAuth Config to enable/disable signature validation for Request Objects.
- Improve the signature validation logic to consider the option explained in the above point.

- After reviewing

### Follow up actions
- Come up with a migration plan for current service providers having their Public certs in their respective
tenant key stores.
- Update documentation.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
@@ -2068,10 +2068,10 @@ public class OAuth2Util {
 
         try {
             ServiceProvider serviceProvider = OAuth2Util.getServiceProvider(clientId, tenantDomain);
-            // Get the certificate content
+            // Get the certificate content.
             String certificateContent = serviceProvider.getCertificateContent();
             if (StringUtils.isNotBlank(certificateContent)) {
-                // Get the cert and return
+                // Build the Certificate object from cert content.
                 return IdentityUtil.convertPEMEncodedContentToCertificate(certificateContent);
             } else {
                 throw new IdentityOAuth2Exception("Public certificate not configured for Service Provider with " +
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
@@ -45,6 +45,10 @@ import org.json.JSONException;
 import org.json.JSONObject;
 import org.wso2.carbon.core.util.KeyStoreManager;
 import org.wso2.carbon.identity.application.authentication.framework.model.AuthenticatedUser;
+import org.wso2.carbon.identity.application.common.IdentityApplicationManagementException;
+import org.wso2.carbon.identity.application.common.model.ServiceProvider;
+import org.wso2.carbon.identity.application.common.util.IdentityApplicationConstants;
+import org.wso2.carbon.identity.application.mgt.ApplicationManagementService;
 import org.wso2.carbon.identity.base.IdentityConstants;
 import org.wso2.carbon.identity.base.IdentityException;
 import org.wso2.carbon.identity.core.util.IdentityConfigParser;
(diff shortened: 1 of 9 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -2046,5 +2050,37 @@ public static boolean isFederatedUser(AuthenticatedUser authenticatedUser) {
...
+            if (StringUtils.isNotBlank(certificateContent)) {
+                // Build the Certificate object from cert content.
+                return IdentityUtil.convertPEMEncodedContentToCertificate(certificateContent);
+            } else {
+                throw new IdentityOAuth2Exception("Public certificate not configured for Service Provider with " +
+                        "client_id: " + clientId + " of tenantDomain: " + tenantDomain);
+            }
+        } catch (CertificateException e) {
+            throw new IdentityOAuth2Exception("Error while building X509 cert of oauth app with client_id: "
+                    + clientId + " of tenantDomain: " + tenantDomain);
```
pulasthi7: exception object is not thrown

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/RequestObjectValidatorImpl.java
```diff
@@ -199,55 +205,21 @@ protected boolean validateAudience(String currentAudience, List<String> audience
...
-    protected Certificate getCertificateForAlias(String tenantDomain, String alias) throws RequestObjectException {
-
-        int tenantId;
-        String error = "Unable to Validate the Signature of Request Object";
-        tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
-        KeyStoreManager keyStoreManager;
-        // get an instance of the corresponding Key Store Manager instance
-        keyStoreManager = KeyStoreManager.getInstance(tenantId);
-        KeyStore keyStore;
+    protected Certificate getPublicCertOfOAuthApp(String clientId, String tenantDomain) throws RequestObjectException {
```
pulasthi7: since certificate is public, "Public" word seems redundant here

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/RequestObjectValidatorImpl.java
```diff
@@ -199,55 +205,21 @@ protected boolean validateAudience(String currentAudience, List<String> audience
...
-            log.error(errorMsg, e);
-            throw new RequestObjectException(OAuth2ErrorCodes.SERVER_ERROR, error);
-        } catch (Exception e) {
-            //keyStoreManager throws Exception
-            log.error("Unable to load key store manager for the tenant domain:" + tenantDomain, e);
-            throw new RequestObjectException(OAuth2ErrorCodes.SERVER_ERROR, error);
+            return OAuth2Util.getPublicCertOfOAuthApp(clientId, tenantDomain);
+        } catch (IdentityOAuth2Exception e) {
+            throw new RequestObjectException("Error retrieving public certificate of OAuth app with client_id: "
+                    + clientId + " of tenantDomain: " + tenantDomain);
```
pulasthi7: e is not thrown

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/RequestObjectValidatorImpl.java
pulasthi7: Also is it safe to remove/rename this method? since this is a protected method there can be extensions.

mefarazath: The idea behind removing this method was that it will no longer be useful since we are moving away from keystore based approach for storing certs. Anyways I agree the right way is to first deprecate and remove the protected method. 

Will reintroduce the method and deprecate it.

PR #807, opened 2018-02-16
Add encryption and hashing support for oauth2 keys

denuwanthi:

This PR will change the oauth2 access tokens, refresh tokens and authorization codes handling as follow:
1.When storing them in database store both the original value/encrypted value and hashed value
2.When searching use the hashed value in the query
3.When decrypting use the encrypted value

This was mainly introduced due to the introduction of dynamic encryption algorithm(RSA + OAEP).
When oauth2 encryption is enabled, we store the encrypted values in the database. In such cases, since the encryption algorithm is dynamic we cannot do search operation using the encrypted value.
Thus we introduce a separate hash column. Since hash value is always same for particular entry we can search by hash and decrypt using encrypted value (since hash is one way cannot decrypt it).

-Introduce new config to set the preferred hashing algorithm
Ex :
<HashAlgorithm>SHA-256</HashAlgorithm>
default value is SHA-256
-Remove queries where consumer secret was used to search with an and operation with consumer key. That is because consumer key is already defined as unique for consumer apps table. Therefore no  need to pass consumer secret 
Update the SQL queries to support that.
-Introduce a new TokenPersistenceProcessor class called HashingPersistenceProcessor to handle hashing before storing the oauth2 keys.
-Provide a method in AbstractOauthDAO to initalize and get the HashingPersistenceProcessor instances.
-Update the AccessTokenDAOImpl class to store hashed token value when doing insert token operations and search using hash value 
-Update the AuthorizationCodeDAOImpl to insert hashed authorization code values search using hashed authorization code values
-Update the SQL query constant class to support new hash columns
-Update TokenMgtDAOImpl to use hashed refresh token value for search operation and  fixed a place where consumer secret was not encrypted.
-Updated the TokenMgtDAO class
-Updated the existing test cases to support newly introduced columns.

This PR should be merged after kernel 4.4.23 is released with following PRs merged
[1]wso2/carbon-kernel#1698
[2]wso2/carbon-kernel#1697

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -695,6 +699,11 @@ public boolean isRefreshTokenRenewalEnabled() {
         return supportedResponseTypes;
     }
 
+    public String getHashAlgorithm() {
+
```
lahirus: empty line unnecessary

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AbstractOAuthDAO.java
```diff
@@ -65,6 +67,15 @@ protected TokenPersistenceProcessor getPersistenceProcessor() {
         return persistenceProcessor;
     }
 
+    /**
+     * Method to get HashingPersistenceProcessor instance
+     * @return
```
lahirus: remove the line if return is not described.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -2068,6 +2077,18 @@ private void parseOpenIDConnectConfig(OMElement oauthConfigElem) {
...
 
+    private void parseHashAlgorithm(OMElement oauthConfigElem) {
+
+        OMElement hashingAlgorithmElement = oauthConfigElem
+                .getFirstChildWithName(getQNameWithIdentityNS(ConfigElements.HASH_ALGORITHM));
+        if (hashingAlgorithmElement != null) {
+            hashAlgorithm = hashingAlgorithmElement.getText();
+        }
+        if (log.isDebugEnabled()) {
+            log.debug("Hash algorithm was set to : " + hashAlgorithm);
```
hasinthaindrajee: Debug log should move to above if block. Else it's default value

denuwanthi: In the begining of the class the hashAlgorithm variable is set to a default value[1].
Therefore it will always print a value 

https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/807/files#diff-fb1b0d809f84709ea4e8a248eddbc573R211

PR #808, opened 2018-02-19
Add scope validator for application management core

senthalan:

PR #809, opened 2018-02-19
add and edit OAuth app with scope validator test changes

senthalan:
depend on #808


senthalan:
merged with implementation of  add scope for app

PR #839, opened 2018-03-06
Remove user consent based claim filtering for back channel grant types

mefarazath:
- $subject

Changes
- Introduced a util method in OIDCClaimUtil to handle filtering claims based on consent. This will avoid code duplication for id_token and user info response building logic.
- In the above-introduced util method added a check to avoid filtering claims based on user consent for back channel grant types. Therefore for back channel grant types such as password, jwt-bearer, saml bearer filtering user claims based on consent will be skipped.
- We have introduced a config in https://github.com/wso2/carbon-identity-framework/pull/1437 to get the list of grant types that require claim filtering based on user consent. Basically, these are the grant type that has an option to prompt the user for consent via a consent page at the moment.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
@@ -20,7 +20,6 @@ package org.wso2.carbon.identity.oauth.config;
 
 import org.apache.axiom.om.OMElement;
 import org.apache.axis2.util.JavaUtils;
-import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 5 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -1743,6 +1752,41 @@ private void parseSupportedGrantTypesConfig(OMElement oauthConfigElem) {
...
+                        .getFirstChildWithName(getQNameWithIdentityNS(ConfigElements.USER_CONSENT_ENABLED_GRANT_TYPE_NAME));
+                String grantTypeName = null;
+                if (grantTypeNameElement != null) {
+                    grantTypeName = grantTypeNameElement.getText();
+                }
+
+                if (StringUtils.isNotEmpty(grantTypeName)) {
+                    userConsentEnabledGrantTypes.add(grantTypeName);
+                } else {
+                    log.warn("Grant Type: " + grantTypeName + " is not a supported grant type. Therefore skipping it " +
```
omindu: This is a bit misleading isn't it ? Since we are only checking for empty

mefarazath: Agreed.

PR #846, opened 2018-03-08
Fix http client issue which fail the user consent

pulasthi7:
Fix http client issue which fail the user consent

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/pom.xml
@@ -156,10 +156,16 @@
             <artifactId>h2</artifactId>
             <scope>test</scope>
         </dependency>
+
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/pom.xml
```diff
@@ -159,7 +159,10 @@
         <dependency>
             <groupId>org.wso2.orbit.org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
-            <scope>provided</scope>
+        </dependency>
+        <dependency>
```
hasinthaindrajee: Shall we add a comment saying that we need to explicitly pack these two jars inside webapp itself. Otherwise this may be removed by someone accidently

pulasthi7: Done.

PR #869, opened 2018-03-29
Fix unable to get country claim in ID token

emswbandara:
Fixes https://github.com/wso2/product-is/issues/3018

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectClaimFilterImpl.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectClaimFilterImpl.java
@@ -194,7 +194,7 @@ public class OpenIDConnectClaimFilterImpl implements OpenIDConnectClaimFilter {
         Map<String, Object> filteredClaims = new HashMap<>();
         List<String> claimUrisInRequestedScope = getClaimUrisInSupportedOidcScope(oidcScopeProperties, oidcScope);
         for (String scopeClaim : claimUrisInRequestedScope) {
-            String oidcClaimUri = getOIDCClaimUri(scopeClaim);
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectClaimFilterImpl.java
```diff
@@ -194,7 +194,7 @@ private Properties getOIDCScopeProperties(String spTenantDomain) {
         Map<String, Object> filteredClaims = new HashMap<>();
         List<String> claimUrisInRequestedScope = getClaimUrisInSupportedOidcScope(oidcScopeProperties, oidcScope);
         for (String scopeClaim : claimUrisInRequestedScope) {
-            String oidcClaimUri = getOIDCClaimUri(scopeClaim);
+            String oidcClaimUri = getOIDCClaimUri(scopeClaim).trim().replaceAll("\n", "");
```
pulasthi7: is it only the newlines causing the issue? Are other whitespace characters ok?

mefarazath: I think we should trim for whitespaces too. Btw we shouldn't have whitespaces or new lines in the first place.

If the oidc property file is correctly populated we shouldn't have to do the trimming here.. Can we check why we get a new line character here in the first place?

nilasini: Seems like it is related to the fix https://github.com/wso2/product-is/issues/2352. 
If we remove the white spaces in the scope's claims defined in the registry then we could able to get the claims as we expected.

emswbandara:
Closing since this is already fixed with https://github.com/wso2/product-is/issues/2352

PR #893, opened 2018-06-01
Add support for adding scope and generate expiry time based on incoming assertion

Megala21:
- This PR provides support for adding scope with self contained access token and creates expiry time based on incoming expiry time assertion.
- Related issue https://github.com/wso2/product-is/issues/3183
- Related PR https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/892

N/A

### Follow up actions
N/A

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
@@ -385,11 +385,9 @@ public class JWTTokenIssuer extends OauthTokenIssuerImpl {
         AuthenticatedUser user;
         long accessTokenLifeTimeInMillis;
         if (authAuthzReqMessageContext != null) {
-            user = authAuthzReqMessageContext.getAuthorizationReqDTO().getUser();
             accessTokenLifeTimeInMillis =
                     getAccessTokenLifeTimeInMillis(authAuthzReqMessageContext, oAuthAppDO, consumerKey);
         } else {
-            user = tokenReqMessageContext.getAuthorizedUser();
             accessTokenLifeTimeInMillis =
                     getAccessTokenLifeTimeInMillis(tokenReqMessageContext, oAuthAppDO, consumerKey);
         }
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
```diff
@@ -416,6 +427,64 @@ protected JWTClaimsSet createJWTClaimSet(OAuthAuthzReqMessageContext authAuthzRe
...
+        String scopeString = null;
+        if (tokenReqMessageContext != null) {
+            scope = tokenReqMessageContext.getScope();
+        } else {
+            scope = authAuthzReqMessageContext.getApprovedScope();
+        }
+        if (ArrayUtils.isNotEmpty(scope)) {
+            scopeString = OAuth2Util.buildScopeString(scope);
+            if (log.isDebugEnabled()) {
+                log.debug("Scope exist for the jwt access token with subject " + jwtClaimsSet.getSubject() + " and "
```
mefarazath: Why not get the subject from the context?
We are building the jwt claimset multiple times unnecessarily just to get context info for logs. IMO we can use information from the context instead

Megala21: Changed with the commit 63aa142358c853c194a1d0035fb99fcaf16ac479

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
```diff
@@ -416,6 +427,81 @@ protected JWTClaimsSet createJWTClaimSet(OAuthAuthzReqMessageContext authAuthzRe
...
+    private String getAuthenticatedSubjectIdentifier(OAuthAuthzReqMessageContext authAuthzReqMessageContext,
+            OAuthTokenReqMessageContext tokenReqMessageContext) {
+
+        AuthenticatedUser authenticatedUser;
+        if (authAuthzReqMessageContext != null) {
+            authenticatedUser = authAuthzReqMessageContext.getAuthorizationReqDTO().getUser();
+        } else {
+            authenticatedUser = tokenReqMessageContext.getAuthorizedUser();
+        }
+        return authenticatedUser != null ? authenticatedUser.getAuthenticatedSubjectIdentifier() : null;
```
mefarazath: In the null case shouldn't we break the flow?
We need to decide whether a self-contained access token with no subject confirmation is acceptable.

Megala21: The relevant authenticated user will be set by relevant grant handlers. Hence the resposibility is purely dependant on the grant handlers. Further I moved this code to separate method from previously existing code segment.

Megala21: Throwing the exception in the event authenticated user is null

PR #918, opened 2018-07-26
Move the OIDC discovery endpoint under the root <issuer>/.well-known/openid-configuration

DMHP:
According to the spec [1] the discovery endpoint should be located in the path /.well-known/openid-configuration.

OpenID Providers supporting Discovery MUST make a JSON document available at the path formed by concatenating the string /.well-known/openid-configuration to the Issuer.

Currently in IS server, the document is located in "oidcdiscovery/.well-known/openid-configuration".

[1] [https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery]

From this PR moved the OIDC discovery endpoint under the root <issuer>/.well-known/openid-configuration

Immediately

### Follow up actions
None


DMHP:
resolves https://github.com/wso2/product-is/issues/3480

PR #957, opened 2018-09-18
Fix Permissions not properly checked in Admin console UI menus

isharailanga:

This PR is to fix https://github.com/wso2/product-is/issues/3747 issue by defining permissions for the OIDC Scopes

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.ui/src/main/resources/META-INF/component.xml
+++ b/components/org.wso2.carbon.identity.oauth.ui/src/main/resources/META-INF/component.xml
@@ -59,7 +59,9 @@
             <order>1</order>
             <style-class>manage</style-class>
             <icon>../oauth/images/add.gif</icon>
-            <require-permission>/permission/admin/manage/identity/claimmgt/metadata/view</require-permission>
+            <require-permission>/permission/admin/manage/identity/claimmgt/metadata/view,
+                /permission/admin/manage/identity/applicationmgt
+            </require-permission>
         </menu>
         <menu>
             <id>list_scope_menu</id>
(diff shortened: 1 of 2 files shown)


darshanasbg:
@isharailanga @nuwandiw @thanujalk 
Shouldn't this UI\Service have its own permission, rather re-using permissions in application-mgt or claim-mgt?

nuwandiw:
> @isharailanga @nuwandiw @thanujalk
> Shouldn't this UI\Service have its own permission, rather re-using permissions in application-mgt or claim-mgt?

The permissions are set according to the services used by these ui operations. 
1. List scopes - uses OAuthAdminService which needs /permission/admin/manage/identity/applicationmgt permissions

2. Add scope - uses claim metadata admin service which needs permission/admin/manage/identity/claimmgt/metadata permission.

So even though we define new permission to above ui component, user wouldn't be able to perform actions without above mentioned permissions.

components/org.wso2.carbon.identity.oauth.ui/src/main/resources/META-INF/component.xml
```diff
@@ -57,6 +59,7 @@
             <order>1</order>
             <style-class>manage</style-class>
             <icon>../oauth/images/add.gif</icon>
+            <require-permission>/permission/admin/manage/identity/claimmgt/metadata/view</require-permission>
```
nuwandiw: This should include /permission/admin/manage/identity/applicationmgt permission as well since adding scopes should be permitted by the OAuthAdminService

darshanasbg:
@nuwandiw Shouldn't it have the both new+required permissions?

nuwandiw:
> @nuwandiw Shouldn't it have the both new+required permissions?

There will not be any new permissions since, **Add** and **List** scope functionalities belong to the OAuthAdminService which already has required permissions defined as /permission/admin/manage/identity/applicationmgt.

darshanasbg:
Ok.. Considering this is just an sub functionality of OAuthAdminService, its ok to re-use permissions that is used for the CRUD operations in the OAuthAdminService

PR #973, opened 2018-11-01
Fix client secret is set to empty after click cancel in OAuth configs

tharindu-bandara:

- Properly update isHashdisabled status to include the cancel button onclick event as well.

#### Documentation

N/A


pulasthi7:
Can you enable hashing for client secret, and verify the secret is shown at regeneration?

tharindu-bandara:
I have tested and verified this fix in hashing enabled client secret regeneration.

PR #1017, opened 2019-02-01
Fix the issue of Token refresh with authorization_code grant type tries to clear cache of invalid token

GDRDABARERA:
Resolves https://github.com/wso2/product-is/issues/4366


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
```diff
@@ -67,7 +67,7 @@
     public static final String DEACTIVATED_ACCESS_TOKEN = "DeactivatedAccessToken";
     private static Log log = LogFactory.getLog(RefreshGrantHandler.class);
     private boolean isHashDisabled = OAuth2Util.isHashDisabled();
-
+
```
Kavindu-Dodan: @GDRDABARERA may be an unwanted addition/style format :) But shouldn't block from getting merged

PR #1030, opened 2019-03-08
Fix updated claims through a custom OIDCClaimsCallbackHandler not reflected in ID token in implicit grant

DMHP:

Fix updated claims through a custom OIDCClaimsCallbackHandler not reflected in ID token in implicit grant

Immediately

### Follow up actions

None


DMHP:
resolves product-is/issues/4060

PR #1055, opened 2019-04-02
Add SP configured audience to self contained access token

nilasini:

Resolves https://github.com/wso2/product-is/issues/4669

We already get the audience values configured through service provider as the audience claims for id token but self contained access token was giving only the client id as the audience value and not considering the audience restrictions we have configured through service provider.

Hence with this fix, I have move the audience retrieval from the service provider configuration inside a Util method and invoked that method in both places (id token builder, JWT token issuer)

can merge this pull request.]

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/cache/OIDCAudienceCache.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/cache/OIDCAudienceCache.java
@@ -17,9 +17,14 @@
 package org.wso2.carbon.identity.oauth.cache;
 
 import org.wso2.carbon.identity.application.common.cache.BaseCache;
+import org.wso2.carbon.identity.oauth.dao.OAuthAppDO;
 import org.wso2.carbon.identity.oauth.listener.OAuthCacheRemoveListener;
 import org.wso2.carbon.utils.CarbonUtils;
 
+/**
+ * @deprecated use {@link OAuthAppDO} to retrieve the audience. Then no need to consider caching for audience.
+ */
+@Deprecated
 public class OIDCAudienceCache extends BaseCache<OIDCAudienceCacheKey, CacheEntry> {
 
     private static final String OAUTH_CACHE_NAME = "OAuthCache";
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
@@ -129,6 +129,7 @@ import java.sql.Timestamp;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -159,6 +160,7 @@ public class OAuth2Util {
     public static final String OPENID_CONNECT = "OpenIDConnect";
     public static final String ENABLE_OPENID_CONNECT_AUDIENCES = "EnableAudiences";
     public static final String OPENID_CONNECT_AUDIENCE = "audience";
+    private static final String OPENID_CONNECT_AUDIENCES = "Audiences";
     private static final String DOT_SEPARATER = ".";
     private static final String IDP_ENTITY_ID = "IdPEntityId";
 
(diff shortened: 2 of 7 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1505,6 +1510,93 @@ public static SpOAuth2ExpiryTimeConfiguration getSpTokenExpiryTimeConfig(String
...
+        return oidcAudiences;
+    }
+
+    private static List<String> getDefinedCustomOIDCAudiences(String clientId, String tenantDomain) {
+
+        List<String> audiences = new ArrayList<>();
+
+        // Priority should be given to service provider specific audiences over globally configured ones.
+        if (OAuth2ServiceComponentHolder.isAudienceEnabled()) {
+            OIDCAudienceCacheKey cacheKey = new OIDCAudienceCacheKey(clientId + "@" + tenantDomain);
```
mefarazath: client_id is unique across tenant domains. So should we use tenantDomain in the key?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1505,6 +1510,93 @@ public static SpOAuth2ExpiryTimeConfiguration getSpTokenExpiryTimeConfig(String
...
+            CacheEntry result = OIDCAudienceCache.getInstance().getValueFromCache(cacheKey);
+
+            // Check cache hit, do the type check.
+            if (result != null && result instanceof OIDCAudienceCacheEntry) {
+                audiences = ((OIDCAudienceCacheEntry) result).getAudiences();
+                return audiences;
+            } else {
+                OAuthAppDAO oAuthAppDAO = new OAuthAppDAO();
+                try {
+                    audiences = oAuthAppDAO.getOIDCAudiences(tenantDomain, clientId);
```
mefarazath: Can't we get the appDO and get the audience from that?
The reason why I propose this that, this will be an additional DB call that we can avoid. 
If we use the method in OAuth2Util to fetch the OAuth app, then we can take advantage of caching as well.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1505,6 +1510,93 @@ public static SpOAuth2ExpiryTimeConfiguration getSpTokenExpiryTimeConfig(String
...
+                    if (CollectionUtils.isNotEmpty(audiences)) {
+                        OIDCAudienceCacheEntry cacheEntry = new OIDCAudienceCacheEntry();
+                        cacheEntry.setAudiences(audiences);
+                        OIDCAudienceCache.getInstance().addToCache(cacheKey, cacheEntry);
```
mefarazath: Why do we need a separate cache for OIDC Audience?
Can't we use the OAuthAppDO to fetch the OIDC audience?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1505,6 +1510,93 @@ public static SpOAuth2ExpiryTimeConfiguration getSpTokenExpiryTimeConfig(String
...
+                        if (log.isDebugEnabled()) {
+                            log.debug("OIDC Audiences added back to the cache.");
```
mefarazath: Needs some context in the logs

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
madurangasiriwardena: @mefarazath this seems to be a refactored method which was originally in DefaultIDTokenBuilder.java. So let's keep that part as it is.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
madurangasiriwardena: All the places that use the `getOIDCAudience` already have the `OAuthAppDO` already retrieved. So passing that object to the `getOIDCAudience` and then using it from `getDefinedCustomOIDCAudiences` would be a better way without retrieving it again.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
madurangasiriwardena: Yes, +1 for that. Check whether there is any specific reason to do this.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/cache/OIDCAudienceCache.java
```diff
@@ -1,45 +0,0 @@
...
- *  limitations under the License.
- */
-
-package org.wso2.carbon.identity.oauth.cache;
-
-import org.wso2.carbon.identity.application.common.cache.BaseCache;
-import org.wso2.carbon.identity.oauth.listener.OAuthCacheRemoveListener;
-import org.wso2.carbon.utils.CarbonUtils;
-
-public class OIDCAudienceCache extends BaseCache<OIDCAudienceCacheKey, CacheEntry> {
```
mefarazath: Let's deprecate the class first and when we remove the class we need to do a major version bump.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1505,6 +1507,75 @@ public static SpOAuth2ExpiryTimeConfiguration getSpTokenExpiryTimeConfig(String
...
+
+        List<String> audiences = new ArrayList<>();
+
+        // Priority should be given to service provider specific audiences over globally configured ones.
+        if (OAuth2ServiceComponentHolder.isAudienceEnabled()) {
+            List<String> audienceList = Arrays.asList(oAuthAppDO.getAudiences());
+            audiences = new ArrayList<>(audienceList);
+            if (CollectionUtils.isNotEmpty(audiences)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("OIDC Audiences " + audiences + " had been retrieved.");
```
mefarazath: can we add the client_id in the log message?

mefarazath:
@nilasini Let's suggest the alternatives for deprecated classes as a doc comment.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1505,6 +1507,76 @@ public static SpOAuth2ExpiryTimeConfiguration getSpTokenExpiryTimeConfig(String
...
+    }
+
+    private static List<String> getDefinedCustomOIDCAudiences(OAuthAppDO oAuthAppDO) {
+
+        List<String> audiences = new ArrayList<>();
+
+        // Priority should be given to service provider specific audiences over globally configured ones.
+        if (OAuth2ServiceComponentHolder.isAudienceEnabled()) {
+            List<String> audienceList = Arrays.asList(oAuthAppDO.getAudiences());
+            audiences = new ArrayList<>(audienceList);
```
sarubi: Why do we have this again since the audienceList is String list?

sarubi: Answer: 
since we want to add one element to the audiences list in the method `getOIDCAudience(String clientId, OAuthAppDO oAuthAppDO)`
 else it will throw "java.lang.UnsupportedOperationException" refer:
https://stackoverflow.com/questions/5755477/java-list-add-unsupportedoperationexception

PR #1060, opened 2019-04-04
Fix tenant oidc logout fails with 'ID token signature validation failed.' error

emswbandara:
Fixes https://github.com/wso2/product-is/issues/4722 and https://github.com/wso2/product-is/issues/4719

**Description**
In OIDC logout we are trying to get tenant domain from subject claim in the id token hint [1] when validating the ID token. This will only work if 'append tenant domain to subject identifier' is selected in the SP configuration. In other scenarios it fails with the error "access_denied ID token signature validation failed." because if subject does not contain the tenant domain, we try to validate the id token with super tenant keystore.

Also this fails when subject identifier is set as email claim, and email contains a different domain such as sathya@gmail.com

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java#L331

**Solution**
As a possible solution, we can get the tenant domain as a request parameter similar to /authorize and /token endpoint implementations instead of relying on the subject claim in the id token.

### DIFF
--- a/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/OIDCSessionConstants.java
+++ b/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/OIDCSessionConstants.java
@@ -31,6 +31,7 @@ public class OIDCSessionConstants {
     public static final String OIDC_POST_LOGOUT_REDIRECT_URI_PARAM = "post_logout_redirect_uri";
     public static final String OIDC_STATE_PARAM = "state";
     public static final String OIDC_SESSION_DATA_KEY_PARAM = "sessionDataKey";
+    public static final String TENANT_DOMAIN = "tenantDomain";
 
     public static final String OIDC_CACHE_CLIENT_ID_PARAM = "client_id";
 
--- a/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
+++ b/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
@@ -219,10 +219,11 @@ public class OIDCLogoutServlet extends HttpServlet {
                 .getParameter(OIDCSessionConstants.OIDC_POST_LOGOUT_REDIRECT_URI_PARAM);
         String state = request
                 .getParameter(OIDCSessionConstants.OIDC_STATE_PARAM);
+        String tenantDomain = request.getParameter(OIDCSessionConstants.TENANT_DOMAIN);
 
         String clientId;
         try {
-            if (!validateIdToken(idTokenHint)) {
+            if (!validateIdToken(idTokenHint, tenantDomain)) {
                 String msg = "ID token signature validation failed.";
                 log.error(msg);
                 redirectURL = OIDCSessionManagementUtil
@@ -274,24 +275,25 @@ public class OIDCLogoutServlet extends HttpServlet {
     /**
      * Validate Id token signature
      * @param idToken Id token
+     * @param tenantDomain tenant Domain
      * @return validation state
      */
-    private boolean validateIdToken(String idToken) {
+    private boolean validateIdToken(String idToken, String tenantDomain) {
 
-        String tenantDomain = getTenantDomainForSignatureValidation(idToken);
-        if (StringUtils.isEmpty(tenantDomain)) {
+        String tenantDomainForSignatureValidation = getTenantDomainForSignatureValidation(idToken, tenantDomain);
+        if (StringUtils.isEmpty(tenantDomainForSignatureValidation)) {
             return false;
         }
-        int tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
+        int tenantId = IdentityTenantUtil.getTenantId(tenantDomainForSignatureValidation);
         RSAPublicKey publicKey;
 
         try {
             KeyStoreManager keyStoreManager = KeyStoreManager.getInstance(tenantId);
 
-            if (!tenantDomain.equals(MultitenantConstants.SUPER_TENANT_DOMAIN_NAME)) {
-                String ksName = tenantDomain.trim().replace(".", "-");
+            if (!MultitenantConstants.SUPER_TENANT_DOMAIN_NAME.equals(tenantDomainForSignatureValidation)) {
+                String ksName = tenantDomainForSignatureValidation.trim().replace(".", "-");
                 String jksName = ksName + ".jks";
-                publicKey = (RSAPublicKey) keyStoreManager.getKeyStore(jksName).getCertificate(tenantDomain)
+                publicKey = (RSAPublicKey) keyStoreManager.getKeyStore(jksName).getCertificate(tenantDomainForSignatureValidation)
                         .getPublicKey();
             } else {
                 publicKey = (RSAPublicKey) keyStoreManager.getDefaultPublicKey();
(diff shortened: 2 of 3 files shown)


components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
```diff
@@ -219,10 +219,11 @@ private String processLogoutRequest(HttpServletRequest request, HttpServletRespo
                 .getParameter(OIDCSessionConstants.OIDC_POST_LOGOUT_REDIRECT_URI_PARAM);
         String state = request
                 .getParameter(OIDCSessionConstants.OIDC_STATE_PARAM);
+        String tenantDomain = request.getParameter(OIDCSessionConstants.TENANT_DOMAIN);
```
mefarazath: do we need to rely on the tenantDomain query param?
Can't we derive the spTenantDomain from the clien_id?

emswbandara: we already do this if 'isJWTSignedWithSPKey' property is set to true [1]. Otherwise we need to get user's tenant domain

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java#L325

components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
```diff
@@ -274,24 +275,25 @@ private String processLogoutRequest(HttpServletRequest request, HttpServletRespo
...
-        int tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
+        int tenantId = IdentityTenantUtil.getTenantId(tenantDomainForSignatureValidation);
         RSAPublicKey publicKey;
 
         try {
             KeyStoreManager keyStoreManager = KeyStoreManager.getInstance(tenantId);
 
-            if (!tenantDomain.equals(MultitenantConstants.SUPER_TENANT_DOMAIN_NAME)) {
-                String ksName = tenantDomain.trim().replace(".", "-");
+            if (!tenantDomainForSignatureValidation.equals(MultitenantConstants.SUPER_TENANT_DOMAIN_NAME)) {
```
DMHP: can tenantDomainForSignatureValidation be null. Shall we swap the constant with variable?

emswbandara: fixed

components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/OIDCSessionConstants.java
```diff
@@ -31,6 +31,7 @@
     public static final String OIDC_POST_LOGOUT_REDIRECT_URI_PARAM = "post_logout_redirect_uri";
     public static final String OIDC_STATE_PARAM = "state";
     public static final String OIDC_SESSION_DATA_KEY_PARAM = "sessionDataKey";
+    public static final String TENANT_DOMAIN = "tenantDomain";
```
mefarazath: tenantDomain is abstract, shouldn't we specifically say which tenantDomain we are looking for?

components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
```diff
@@ -274,24 +275,25 @@ private String processLogoutRequest(HttpServletRequest request, HttpServletRespo
...
-        int tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
+        int tenantId = IdentityTenantUtil.getTenantId(tenantDomainForSignatureValidation);
         RSAPublicKey publicKey;
 
         try {
             KeyStoreManager keyStoreManager = KeyStoreManager.getInstance(tenantId);
 
-            if (!tenantDomain.equals(MultitenantConstants.SUPER_TENANT_DOMAIN_NAME)) {
-                String ksName = tenantDomain.trim().replace(".", "-");
+            if (!MultitenantConstants.SUPER_TENANT_DOMAIN_NAME.equals(tenantDomainForSignatureValidation)) {
```
mefarazath: Shall we extract the validating key resolving logic to separate method?

components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
```diff
@@ -314,21 +316,28 @@ private boolean validateIdToken(String idToken) {
...
             String clientId = extractClientFromIdToken(idToken);
             if (isJWTSignedWithSPKey) {
                 OAuthAppDO oAuthAppDO = OAuth2Util.getAppInformationByClientId(clientId);
-                tenantDomain = OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO);
+                tenantDomainForSignatureValidation = OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO);
             } else {
-                //It is not sending tenant domain with the subject in id_token by default, So to work this as
-                //expected, need to enable the option "Use tenant domain in local subject identifier" in SP config
-                tenantDomain = MultitenantUtils.getTenantDomain(extractSubjectFromIdToken(idToken));
+                if (StringUtils.isNotBlank(tenantDomain)) {
```
mefarazath: It would be worth to add a few debug logs to indicate which path the request went through.

components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
```diff
@@ -314,21 +316,28 @@ private boolean validateIdToken(String idToken) {
...
+                    // Tenant domain is available in request parameters.
+                    tenantDomainForSignatureValidation = tenantDomain;
+                } else {
+                    //It is not sending tenant domain with the subject in id_token by default, So to work this as
```
mefarazath: A debug logs saying that we are falling back to get tenant domain from subject and we rely on the SP configuration option "Use tenant domain in local subject identifier" would be helpful.

components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/OIDCSessionConstants.java
emswbandara: This is the name of the request parameter. Same query param name is used with token endpoint and authz endpoint to specify user tenant domain. Therefore I used the same here for consistency.
[1] https://docs.wso2.com/display/IS570/Try+Authorization+Code+Grant

mefarazath: Actually, authorize URL doesn't make use of this tenantDomain parameter. Internally tenantDomin is derived based on the client_id. 
We need to fix the docs I guess.

malithie:
@emswbandara , shall we schedule up a review on this

emswbandara:
Re-initiated via https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1063

PR #1061, opened 2019-04-04
Fix #4711:Incorrect behavior when empty client id is sent in authorize request

IsurangaPerera:

- Resolves wso2/product-is#4711

In the current implementation, app-validity and app-state are checked at the time of building the OAuthMessage object and return a JSON string if clientId is an invalid one. Since the error message is seen by the end-user instead of returning a plain JSON string we will redirect the users to an error page.
This is not against the OAuth 2.0 spec as stated in **section 4.2.2.1**

> If the request fails due to a missing, invalid, or mismatching
>    redirection URI, or if the client identifier is missing or invalid,
>    the authorization server SHOULD inform the resource owner of the
>    error and MUST NOT automatically redirect the user-agent to the
>    invalid redirection URI.
> 
>    If the resource owner denies the access request or if the request
>    fails for reasons other than a missing or invalid redirection URI,
>    the authorization server informs the client by adding the following
>    parameters to the fragment component of the redirection URI using the
>    "application/x-www-form-urlencoded" format, per Appendix B: 
> 
**Test Case changes**
- Add test case to handle empty client_id
- Change invalid client Id test case so that client_id and app is validated at the time of app validation
- In case of invalid/empty client_id, user will be redirected to an error page instead of sending a **401** response

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/state/OAuthRequestStateValidator.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/state/OAuthRequestStateValidator.java
@@ -114,15 +114,6 @@ public class OAuthRequestStateValidator {
                 oAuthMessage.setSessionDataKeyFromConsent(null);
             }
         }
-
-        validateOauthApplication(oAuthMessage);
-    }
-
-    private void validateOauthApplication(OAuthMessage oAuthMessage) throws InvalidRequestParentException {
-
-        if (StringUtils.isNotBlank(oAuthMessage.getClientId())) {
-            EndpointUtil.validateOauthApplication(oAuthMessage.getClientId());
-        }
     }
 
     private void validateInputParameters(OAuthMessage oAuthMessage) throws InvalidRequestException {
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
@@ -22,11 +22,10 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.oltu.oauth2.common.message.types.GrantType;
+import org.owasp.encoder.Encode;
 import org.wso2.carbon.core.AbstractAdmin;
 import org.wso2.carbon.identity.base.IdentityException;
 import org.wso2.carbon.identity.core.util.IdentityTenantUtil;
-import org.wso2.carbon.identity.oauth.IdentityOAuthAdminException;
-import org.wso2.carbon.identity.oauth.OAuthAdminService;
 import org.wso2.carbon.identity.oauth.OAuthUtil;
 import org.wso2.carbon.identity.oauth.cache.CacheEntry;
 import org.wso2.carbon.identity.oauth.cache.OAuthCache;
@@ -74,6 +73,7 @@ import java.util.Map;
 public class OAuth2Service extends AbstractAdmin {
 
     private static Log log = LogFactory.getLog(OAuth2Service.class);
+    private static final String APP_STATE_ACTIVE = "ACTIVE";
 
     /**
      * Process the authorization request and issue an authorization code or access token depending
@@ -130,8 +130,27 @@ public class OAuth2Service extends AbstractAdmin {
         }
 
         try {
-            OAuthAppDAO oAuthAppDAO = new OAuthAppDAO();
-            OAuthAppDO appDO = oAuthAppDAO.getAppInformation(clientId);
+            if (StringUtils.isBlank(clientId)) {
+                throw new InvalidOAuthClientException("Invalid client_id. No OAuth application has been registered " +
+                        "with the given client_id");
+            }
+            OAuthAppDO appDO = OAuth2Util.getAppInformationByClientId(clientId);
+            String appState = appDO.getState();
+
+            if (StringUtils.isEmpty(appState)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("A valid OAuth client could not be found for client_id: " + clientId);
+                }
+                throw new InvalidOAuthClientException("A valid OAuth client could not be found for client_id: " +
+                        Encode.forHtml(clientId));
+            }
+
+            if (!appState.equalsIgnoreCase(APP_STATE_ACTIVE)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("App is not in active state in client ID: " + clientId + ". App state is: " + appState);
+                }
+                throw new InvalidOAuthClientException("Oauth application is not in active state");
+            }
 
             if (StringUtils.isEmpty(appDO.getGrantTypes()) || StringUtils.isEmpty(appDO.getCallbackUrl())) {
                 if (log.isDebugEnabled()) {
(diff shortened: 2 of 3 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -130,9 +130,31 @@ public OAuth2ClientValidationResponseDTO validateClientInfo(String clientId, Str
         }
 
         try {
+            if (StringUtils.isBlank(clientId)) {
+                throw new InvalidOAuthClientException("Invalid client_id. No OAuth application has been registered " +
+                        "with the given client_id");
+            }
             OAuthAppDAO oAuthAppDAO = new OAuthAppDAO();
```
mefarazath: We should avoid calling the DAO directly since it's not a cached back DAO. For the moment we have to rely on using the OAuth2Util.getAppInformationByClientId(clientId) method which makes use of the cache.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -148,7 +170,7 @@ public OAuth2ClientValidationResponseDTO validateClientInfo(String clientId, Str
             OAuth2Util.setClientTenatId(IdentityTenantUtil.getTenantId(appDO.getUser().getTenantDomain()));
 
             // Valid Client, No callback has provided. Use the callback provided during the registration.
-            if (callbackURI == null) {
+            if (StringUtils.isBlank(callbackURI)) {
```
mefarazath: 1. Not sending the 'redirect_uri' query param is allowed in the spec. 
2. If the redirect_uri query param is sent with an empty value we should consider it as an invalid request.

With this change, both cases will be considered valid. This is not the expected behaviour isn't it?

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/state/OAuthRequestStateValidator.java
```diff
@@ -114,15 +114,6 @@ private void validateRequest(OAuthMessage oAuthMessage)
                 oAuthMessage.setSessionDataKeyFromConsent(null);
             }
         }
-
-        validateOauthApplication(oAuthMessage);
-    }
-
-    private void validateOauthApplication(OAuthMessage oAuthMessage) throws InvalidRequestParentException {
```
mefarazath: Any reason why we removed this method?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
mefarazath: The same class has getOauthApplicationState(String consumerKey) method which does this. Let's reuse that here.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
IsurangaPerera: Fixed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -130,8 +130,29 @@ public OAuth2ClientValidationResponseDTO validateClientInfo(String clientId, Str
...
+            }
+            OAuthAppDO appDO = OAuth2Util.getAppInformationByClientId(clientId);
+
+            String appState = appDO.getState();
+
+            if (StringUtils.isEmpty(appState)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("A valid OAuth client could not be found for client_id: " + clientId);
+                }
+
```
ashensw: Remove unnecessary new lines.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -130,8 +130,29 @@ public OAuth2ClientValidationResponseDTO validateClientInfo(String clientId, Str
...
 
         try {
-            OAuthAppDAO oAuthAppDAO = new OAuthAppDAO();
-            OAuthAppDO appDO = oAuthAppDAO.getAppInformation(clientId);
+            if (StringUtils.isBlank(clientId)) {
+                throw new InvalidOAuthClientException("Invalid client_id. No OAuth application has been registered " +
+                        "with the given client_id");
+            }
+            OAuthAppDO appDO = OAuth2Util.getAppInformationByClientId(clientId);
+
```
ashensw: Remove unnecessary new lines.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -130,8 +130,29 @@ public OAuth2ClientValidationResponseDTO validateClientInfo(String clientId, Str
         }
 
         try {
-            OAuthAppDAO oAuthAppDAO = new OAuthAppDAO();
-            OAuthAppDO appDO = oAuthAppDAO.getAppInformation(clientId);
+            if (StringUtils.isBlank(clientId)) {
+                throw new InvalidOAuthClientException("Invalid client_id. No OAuth application has been registered " +
```
ashensw: Is the error msg is correct?

IsurangaPerera: Since the end-user is redirected to the error page, having a more specific error message such as client_id is null or empty is not necessary.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -130,8 +130,27 @@ public OAuth2ClientValidationResponseDTO validateClientInfo(String clientId, Str
...
+                if (log.isDebugEnabled()) {
+                    log.debug("A valid OAuth client could not be found for client_id: " + clientId);
+                }
+                throw new InvalidOAuthClientException("A valid OAuth client could not be found for client_id: " +
+                        Encode.forHtml(clientId));
+            }
+
+            if (!appState.equalsIgnoreCase(APP_STATE_ACTIVE)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("App is not in active state in client ID: " + clientId + ". App state is:" + appState);
```
ashensw: Add a space after `state is:`

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/state/OAuthRequestStateValidator.java
IsurangaPerera: With the modifications(as stated in the description) we no longer need this validation method as the clientId validation will be done at the time of app validation.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
IsurangaPerera: Fixed

PR #1065, opened 2019-04-08
Fix #4403:Sending access_token through /oauth2/userinfo API Body in WSO2 5.7.0

IsurangaPerera:
- Resolves wso2/product-is#4403
When CorrelationIdValve has enabled it parses the input stream and creates the parameter map. In the current OAuth implementation, it again tries to parse the raw input stream and get parameters hence it fails to get the access token. In this fix, we simply get the access_token parameter by accessing the parameter map(without dealing with the raw input stream.)

- **Test case changes**
Remove testing scenarios related to request body parsing since the request is no longer being parser manually.

- ASAP

### Follow up actions
- N/A

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java
@@ -35,6 +35,7 @@ import org.wso2.carbon.identity.oauth2.dto.OAuth2TokenValidationResponseDTO;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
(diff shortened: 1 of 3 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/impl/UserInforRequestDefaultValidator.java
```diff
@@ -24,81 +24,47 @@
...
-                        return arrAccessTokenNew[0];
-                    }
-                }
-                return arrAccessToken[1];
-            } else {
-                throw new UserInfoEndpointException(OAuthError.ResourceResponse.INVALID_REQUEST,
-                        "Content-Type header is wrong");
-            }
+        String authzHeaders = request.getHeader(HttpHeaders.AUTHORIZATION);
+        String access_token = request.getParameter(ACCESS_TOKEN_PARAM);
```
madurangasiriwardena: shall we rename the variable to camel case?

madurangasiriwardena:
@IsurangaPerera did you verify the scenario where the token is sent in the Authorization header?

IsurangaPerera:
@madurangasiriwardena Yes I manually tested and there are unit tests to cover that scenarios as well

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/user/OpenIDConnectUserEndpoint.java
```diff
@@ -55,6 +56,7 @@
 
     @GET
     @Path("/")
+    @Consumes("application/x-www-form-urlencoded")
```
mefarazath: Need to check whether this would break existing clients. Since we do not expect a content type header with user info response

PR #1084, opened 2019-04-21
Fix #5078:Authenticate to provision a user with OAuth with sufficient privileges fails

IsurangaPerera:
- Fix wso2/product-is#5078
Here we have authentication flow and authorization flow. Token validation service is used at authentication and username is set by the validator. However, when "Use user store domain in local subject identifier" is not enabled in **Local & Outbound Authentication Configuration** usernames of secondary user-store users are set without the domain parameter.
Since the user realm is configured depending on the tenant name when authorizing secondary user store users at Authorization Valve suffering from a caching issue which resulted in permission update ignorance.
This issue is not affected to tenant users as the tenant is explicitly appended to the username at the time of authentication.
To overcome this issue we have appended user-store domain at the time of authentication which will later be used to get the correct user store.

- After merging https://github.com/wso2-extensions/identity-carbon-auth-rest/pull/69

### Follow up actions
- N/A


malithie:
@IsurangaPerera , I don't think this fix is valid here.
Basically, the validation service can access to user related to the token and authorizations should be checked within that user's realm.

IsurangaPerera:
@malithie Please find the updated description.
Using thread-local variables creates additional problems in this case I have user context parameters in requestDTO recognize authentication flow.

PR #1086, opened 2019-04-22
Token Issuer Related Fixes

vihanga-liyanage:
- This is the public fix for https://github.com/wso2/product-is/issues/4913
- The root cause for the issue was the way we treat the tokens in different circumstances. When the token is JWT or similar, we store a representation of the token (ex: JTI) instead of the actual token in DBs. We have not considered this fact in other flows such as token validation. 
- With this PR, we use a matching token issuer in the token validation flow and use its getTokenHash() method to get the actual token.
- To choose the matching token issuer, we go through a chain of available token issuers.


vihanga-liyanage:
Build was ran and successful. https://jenkins-support-all.wso2.com/jenkins/job/DEV_PR_BUILDER/50/

PR #1087, opened 2019-04-23
Fix #5078:Authenticate to provision a user with OAuth with sufficient privileges fails

IsurangaPerera:
- Fix wso2/product-is#5078
Here we have authentication flow and authorization flow. Token validation service is used at authentication and username is set by the validator. However, when "Use user store domain in local subject identifier" is not enabled in **Local & Outbound Authentication Configuration** usernames of secondary user-store users are set without the domain parameter.
Since the user realm is configured depending on the tenant name when authorizing secondary user store users at Authorization Valve suffering from a caching issue which resulted in permission update ignorance.
This issue is not affected to tenant users as the tenant is explicitly appended to the username at the time of authentication.
To overcome this issue we have appended user-store domain at the time of authentication which will later be used to get the correct user store.

- After merging https://github.com/wso2-extensions/identity-carbon-auth-rest/pull/69

### Follow up actions
- N/A


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
```diff
@@ -59,6 +59,7 @@
     private Log log = LogFactory.getLog(TokenValidationHandler.class);
     private Map<String, OAuth2TokenValidator> tokenValidators = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
     private static final String BUILD_FQU_FROM_SP_CONFIG = "OAuth.BuildSubjectIdentifierFromSPConfig";
+    private static final String AUTH_FLOW = "OAuth.AuthFlow";
```
ayshsandu: Can we move this constant to a common location? I see it is repeated in https://github.com/wso2-extensions/identity-carbon-auth-rest/pull/69 as well.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
```diff
@@ -468,14 +471,18 @@ private OAuth2IntrospectionResponseDTO validateAccessToken(OAuth2TokenValidation
...
-    private String getAuthzUser(AccessTokenDO accessTokenDO) throws IdentityOAuth2Exception {
+    private String getAuthzUser(AccessTokenDO accessTokenDO, boolean isAuthFlow) throws IdentityOAuth2Exception {
 
         AuthenticatedUser user = accessTokenDO.getAuthzUser();
 
         if (user.isFederatedUser()) {
             return user.getAuthenticatedSubjectIdentifier();
         }
 
+        if (isAuthFlow) {
```
mefarazath: Why do we need to this?
I guess that deserves a comment in the code

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
```diff
@@ -711,4 +718,11 @@ private boolean isIDTokenEncrypted(String idToken) {
         return StringUtils.countMatches(idToken, ".") == 4;
     }
 
+    private boolean checkIfAuthFlow(OAuth2TokenValidationRequestDTO.TokenValidationContextParam[] context) {
+
+        return context != null && context.length == 1 && StringUtils.isNotBlank(context[0].getKey()) && context[0]
```
mefarazath: What are we checking here?
Where does this value set from?

Better to explain that via a comment

IsurangaPerera: Fixed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
IsurangaPerera: Added comments

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
mefarazath: if there is no special reason lets use the full form "isAuthenticationFlow()" or "isCalledFromAuthenticationAPI" or something along the lines to convey the meaning.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
```diff
@@ -468,14 +471,20 @@ private OAuth2IntrospectionResponseDTO validateAccessToken(OAuth2TokenValidation
...
-    private String getAuthzUser(AccessTokenDO accessTokenDO) throws IdentityOAuth2Exception {
+    private String getAuthzUser(AccessTokenDO accessTokenDO, boolean isAuthFlow) throws IdentityOAuth2Exception {
 
         AuthenticatedUser user = accessTokenDO.getAuthzUser();
 
         if (user.isFederatedUser()) {
             return user.getAuthenticatedSubjectIdentifier();
         }
 
+        //This checks if the request originated from authentication API. If so username is returned with the
```
nilasini: Add a space when starting the sentence

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
IsurangaPerera: Fixed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
IsurangaPerera: Move the constant to OauthConstants and with this modification Oauth version in wso2-extensions/identity-carbon-auth-rest need to be updated to the immediate version which includes this modification

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
```diff
@@ -711,4 +719,14 @@ private boolean isIDTokenEncrypted(String idToken) {
         return StringUtils.countMatches(idToken, ".") == 4;
     }
 
+    private boolean isOriginatedFromAuthenticationAPI(OAuth2TokenValidationRequestDTO
+                                                              .TokenValidationContextParam[] context) {
+
+        // If the origin of the request is from the authentication API "OAuth.AuthFlow" should be set as
```
ayshsandu: The comment might be misleading as "OAuth.AuthFlow" has changed to "OAuth.isAuthenticationFlow"

IsurangaPerera: Fixed

malithie:
@IsurangaPerera , I'm wondering why we need to have a specific property set and to retrieve the user store domain based on that one with respect to this specific flow.
Is the authorization handled in the valve based on the user returned from the introspection endpoint

shouldn't this be handled by a more standard way rather than being coupled for a specific flow.

IsurangaPerera:
@malithie When authenticating using access token authentication valve should exactly determine the username which will be used to uniquely identify the user. In that case, the username should contain the user-store domain and tenant domain appended to it.

But at the moment access token validation uses the method used in introspection. But in introspection, it is not required that username consists of the user-store domain and tenant domain(This can be configured explicitly). So this flow is not valid for authentication & authorization.

Even though username received at token validation may not contain tenant domain it is appended explicitly at OAuthAccessTokenHandler at authentication valve. However, users from secondary user-store are not handled in this way.

 Since the user realm is configured based on tenant domain secondary user-store can only be recognized by user-store domain identifier in username at the Authorization valve.

malithie:
From the fix and your explanation above, user store domain of the user is identified from the introspection response. What I'm saying is a requirement of identifying the user realm can be a very common use case. I'm -1 to coupling that to this respective flow.
In addition, oauth itself is to handle authorization, I'm still trying to understand why we had to get the user from introspection endpoint and had to authorize that user again at a later point. 

As you have said this is the same approach followed, appreciate if you can point such places

PR #1089, opened 2019-04-23
Avoid unnecessary token cache cleanup during OAuth app creation

mefarazath:
During oauth app creation the app owner is assigned a new role (Application/<app_name>). This is reflected to the listeners as a user attribute update.
As a result, IdentityOathEventListener tries to clear cached claims against tokens for the app owner. This is valid in cases where app owners claims are included
in id_tokens and user info response. However, this cache clearing logic is applied for all tokens issued for the particular app owner. Tokens issued for client_credentials
grant type do not usually cache claims against them since id_tokens are not allowed for client_credential grant.

This improves the cache clearing logic to only attempt cache clearing on tokens that could potentially have claims cached against them without blindly clearing the cache for all the tokens

Fixes https://github.com/wso2/product-is/issues/5050


mefarazath:
This is the initial fix to improve the AuthorizationGrantCache clearing logic during an app creation.

During an app creation a role is assigned to the app owner. This role assignment is reflected as a user attribute update. The user attribute update triggers the listener to clear cache entries of AuthorizationGrantCache which contain user attributes.

The idea of this fix is to avoid cache clearing against access tokens issued for client_credentials grant type (unless we explicitly allow id_tokens to be issued for client_credentials grant type). The reason being that if we do not issue id_tokens for client_credentials grant type then we do not store attributes in the AuthorizationGrantCache against those tokens.

However, we can further improve this logic by only considering tokens (tokens issued for users) that contain "openid" scope. This would futher improve the listener and avoid performance degradation.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
```diff
@@ -1104,6 +1131,28 @@ public void revokeAccessToken(String tokenId, String userId) throws IdentityOAut
...
     }
 
+    /**
+     * Checks whether id_tokens are issued for application tokens (ie. tokens issued for client_credentials grant type)
+     *
+     * @return
+     */
+    private boolean isIdTokenIssuedForApplicationTokens() {
+
+        return !OAuthServerConfiguration.getInstance().getIdTokenNotAllowedGrantTypesSet()
```
ruwanta: getIdTokenNotAllowedGrantTypesSet can be null.

mefarazath: getIdTokenNotAllowedGrantTypesSet() returns an empty set if there aren't any values

ruwanta: +1

mefarazath:
> This is the initial fix to improve the AuthorizationGrantCache clearing logic during an app creation.
> 
> During an app creation a role is assigned to the app owner. This role assignment is reflected as a user attribute update. The user attribute update triggers the listener to clear cache entries of AuthorizationGrantCache which contain user attributes.
> 
> The idea of this fix is to avoid cache clearing against access tokens issued for client_credentials grant type (unless we explicitly allow id_tokens to be issued for client_credentials grant type). The reason being that if we do not issue id_tokens for client_credentials grant type then we do not store attributes in the AuthorizationGrantCache against those tokens.
> 
> However, we can further improve this logic by only considering tokens (tokens issued for users) that contain "openid" scope. This would futher improve the listener and avoid performance degradation.

Create https://github.com/wso2/product-is/issues/5126 to track this

PR #1090, opened 2019-04-23
update waffle-jna version

fazlan-nazeem:


pom.xml
```diff
@@ -792,9 +792,9 @@
         <org.apache.oltu.oauth2.client.version>1.0.0</org.apache.oltu.oauth2.client.version>
         <oltu.package.import.version.range>[1.0.0, 2.0.0)</oltu.package.import.version.range>
 
-        <waffle.imp.pkg.version.range>[1.6.wso2v5, 2.0)</waffle.imp.pkg.version.range>
-        <waffle-jna.wso2.version>1.6.wso2v5</waffle-jna.wso2.version>
-        <waffle-jna.imp.pkg.version.range>[1.6.wso2v1, 2.0)</waffle-jna.imp.pkg.version.range>
+        <waffle.imp.pkg.version.range>[1.6.wso2v6, 2.0)</waffle.imp.pkg.version.range>
```
ruwanta: import range needs to be kept unchanged. 
Also the imp range seems odd. Usually import range does not need wso2 suffix, but the something like 1.6.0 (which is the best practice int OSGI import.

fazlan-nazeem: @ruwanta Fixed as suggested.

PR #1111, opened 2019-05-14
Fix cache not updated when client app secret is regenerated

mefarazath:
Fixes https://github.com/wso2/product-is/issues/5240

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminService.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminService.java
@@ -148,7 +148,6 @@ public class OAuthAdminService extends AbstractAdmin {
     public OAuthConsumerAppDTO getOAuthApplicationData(String consumerKey) throws IdentityOAuthAdminException {
 
         OAuthConsumerAppDTO dto;
-        OAuthAppDAO dao = new OAuthAppDAO();
         try {
             OAuthAppDO app = getOAuthApp(consumerKey);
             if (app != null) {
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminService.java
```diff
@@ -154,6 +153,7 @@ public OAuthConsumerAppDTO getOAuthApplicationData(String consumerKey) throws Id
         try {
             OAuthAppDO app = dao.getAppInformation(consumerKey);
             if (app != null) {
+                AppInfoCache.getInstance().addToCache(consumerKey, app);
```
malithie: Is this method being used throughout to retrieve app from db always.

mefarazath: Yes. It seems so.
In fact, none of the App information retrieval within this class make use of the cache. 

Ideally, we should have a cache backed DAO. But for the moment let me improve the code to make use of app info cache before trying to retrieve from DB

malithie: +1

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminService.java
```diff
@@ -571,20 +569,20 @@ public boolean isScopeExist(String scope) throws IdentityOAuthAdminException {
      */
     public void updateConsumerAppState(String consumerKey, String newState) throws IdentityOAuthAdminException {
 
-        OAuthAppDAO oAuthAppDAO = new OAuthAppDAO();
         try {
             OAuthAppDO oAuthAppDO = AppInfoCache.getInstance().getValueFromCache(consumerKey);
             if (oAuthAppDO == null) {
-                oAuthAppDO = oAuthAppDAO.getAppInformation(consumerKey);
+                oAuthAppDO = getOAuthApp(consumerKey);
```
malithie: We can directly use 'getOAuthApp' method being newly implemented right, rather than retrieving from the cache specifically first

mefarazath: Yes :)

mefarazath: Updated the code to use the getOAuthApp method

PR #1118, opened 2019-06-04
Make the error direction compliant with OAuth Spec

pamodaaw:

This PR changes the redirect URL when there is an exception. The behavior is changed to be compliant with the[ OAuth specification](https://tools.ietf.org/html/rfc6749#section-4.1.2.1). To achieve the behavior, a new set of OAuth error codes and a new configuration is introduced. From the next major release onwards this property will be enabled by default.

> - If the new property is **true** OR if the **request fails due to a missing, invalid, or mismatching redirection URI, or if the client identifier is missing or invalid**, 
>     - the authorization server SHOULD inform the resource owner of the error and MUST NOT automatically redirect the user-agent to the invalid redirection URI.
> - If the new property is **false** and if the resource owner denies the access request or if the request fails for reasons other than a missing or invalid redirection URI, 
>    - the authorization server informs the client by adding the error code, error message and state parameters to the query component of the redirection URI.

#### Config Changes
New property - **RedirectToRequestedRedirectUri**
Default value - true
Steps to follow f it is required to redirect the error to the redirect URI given in the request, 
- add 
`<RedirectToRequestedRedirectUri>false</RedirectToRequestedRedirectUri>` under the <OAuth> tag in <CARBON_HOME>/repository/conf/identity/identity.xml file

### Follow up actions

1. Update the migration documents

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuth2ErrorCodes.java
+++ b/components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuth2ErrorCodes.java
@@ -35,4 +35,20 @@ public class OAuth2ErrorCodes {
     private OAuth2ErrorCodes(){
 
     }
+
+    public class OAuth2SubErrorCodes {
+
+        public static final String INVALID_PKCE_CHALLENGE_CODE = "invalid_pkce_challenge_code";
+        public static final String INVALID_CLIENT = "invalid_client";
+        public static final String INVALID_REDIRECT_URI = "invalid_redirect_uri";
+        public static final String SESSION_TIME_OUT = "session_time_out";
+        public static final String CERTIFICATE_ERROR = "certificate_error";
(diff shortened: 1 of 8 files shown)


components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuth2ErrorCodes.java
```diff
@@ -35,4 +35,20 @@
     private OAuth2ErrorCodes(){
 
     }
+
+    public class OAuth2SubErrorCodes {
```
JKAUSHALYA: Shall we have a class level comment explaining what is this?

pamodaaw: Added with the PR https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1144

PR #1123, opened 2019-06-09
Fix - role based scope validation added to implicit grant

vihanga-liyanage:
Resolves wso2/product-is#5335

Apart from fixing the implicit flow, this change will enforce scope validation in authorization code grant type authorization request as well. The previous implementation only validated scopes in token request and a valid code was returned even for an invalid scope. This is a spec violation.


vihanga-liyanage:
The build was successful with the changes. https://jenkins-support-all.wso2.com/jenkins/job/DEV_PR_BUILDER/59/

PR #1173, opened 2019-09-11
Stop handling oidc logout twice

janakamarasena:

When a logout request is triggered from an OIDC app a SESSION_TERMINATE event gets fired and the OIDCLogoutEventHandler also attempts to handle the request. This change stops that.

Resolves: https://github.com/wso2/product-is/issues/6325


components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/handler/OIDCLogoutEventHandler.java
```diff
@@ -49,6 +51,14 @@ public void handleEvent(Event event) throws IdentityEventException {
             Cookie opbsCookie = null;
 
             if (request != null) {
+                if (StringUtils.equals(request.getParameter(COMMON_AUTH_CALLER_PATH),
```
darshanasbg: This will work when logout request only dealing with IS.

If IS need to federate the logout request to a federated IdP, then the `request` object here, will be the logout response from federated IDP, hence it won't have the `COMMON_AUTH_CALLER_PATH` parameter.

Check on the above scenario and improve the fix to work in both cases.

PR #1190, opened 2019-10-11
Add 'state' parameter to the IDP's page URL once logged-out

tharindu-bandara:
- Fixes wso2/product-is#6614.
- The 'state' parameter received in the logout request is dropped, if the id_token_hint parameter is not available. This PR changes this behavior in the above scenario and propagate this value to the IDP's logout page(since id_token_hint parameter is not available) to support a case if IDP's logout page is customized.

#### Documentation

- N/A


tharindu-bandara:
> Requesting changes based on the failed Jenkins job https://jenkins-support-all.wso2.com/jenkins/job/DEV_PR_BUILDER/174/, started by Tharindu Bandara,

One test case failure with DEV_PR_BUILDER. Which fails even without this fix in the Jenkins project. Other tests have passed successfully.

PR #1191, opened 2019-10-14
Fix OIDC hybrid flow not working without Implicit grant configured in SP

DMHP:
OIDC hybrid flow depends on implicit grant type when working with authorization code grant type. This fix improve the code to validate whether the implicit grant is enabled, only if the response types are id_token token, id_token, token

Immediately

### Follow up actions
None


DMHP:
https://github.com/wso2/product-is/issues/6616

PR #1309, opened 2020-02-03
Remove unwanted deprecation and improve styling

vihanga-liyanage:
We've deprecated **getAccessTokensByUser()** and **getAuthorizationCodesByUser()** methods, and  introduced different ones as a fix to the issue https://github.com/wso2/product-is/issues/5126. 

These methods were used in the method **IdentityOathEventListener.removeTokensFromCache()** and the above changes were done thinking of that usage only. But those methods are public methods and there's is no need to deprecate them, just because our default implementation uses different methods.


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/IdentityOathEventListener.java
```diff
@@ -372,6 +372,8 @@ private void removeTokensFromCache(String userName, UserStoreManager userStoreMa
         authenticatedUser.setTenantDomain(tenantDomain);
         authenticatedUser.setUserName(userName);
         try {
+            // Only the tokens and auth codes issued for openid scope should be removed from the cache, since no
+            // claims are usually cached against tokens or auth codes, otherwise.
```
dewniMW: Shall we make this a multi-line comment?

vihanga-liyanage: done

PR #1312, opened 2020-02-07
Allow native mobile app callback URLs in OAuth2

vihanga-liyanage:
Resolves https://github.com/wso2/product-is/issues/6109


mefarazath:
What about the update page? It should have the same logic too right?

vihanga-liyanage:
> What about the update page? It should have the same logic too right?

Fixed.

PR #1328, opened 2020-02-21
Support extension for OIDProviderResponseBuilder

emswbandara:
Fixes wso2/product-is#7533

### DIFF
--- a/components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/DiscoveryConstants.java
+++ b/components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/DiscoveryConstants.java
@@ -54,6 +54,32 @@ public class DiscoveryConstants {
      */
     public static final String REVOCATION_ENDPOINT = "Revocation_endpoint";
     /**
+     * revocation_endpoint_auth_methods_supported
+     * OPTIONAL. JSON array containing a list of Client Authentication methods supported by this Revocation
+     * Endpoint. The options are client_secret_post, client_secret_basic, client_secret_jwt, and
+     * private_key_jwt, as described in Section 9 of OpenID Connect Core 1.0 [OpenID.Core]. Other
+     * authentication methods MAY be defined by extensions. If omitted, the default is client_secret_basic
+     * -- the HTTP Basic Authentication Scheme specified in Section 2.3.1 of OAuth 2.0 [RFC6749].
+     */
+    public static final String REVOCATION_ENDPOINT_AUTH_METHODS_SUPPORTED =
+            "Revocation_endpoint_auth_methods_supported";
+    /**
+     * introspection_endpoint
+     * OPTIONAL. URL of the OP's Introspection Endpoint [OpenID.Core]. This URL MUST use the https scheme and
+     * MAY contain port, path, and query parameter components.
+     */
+    public static final String INTROSPECTION_ENDPOINT = "Introspection_endpoint";
+    /**
+     * introspection_endpoint_auth_methods_supported
+     * OPTIONAL. JSON array containing a list of Client Authentication methods supported by this Introspection
+     * Endpoint. The options are client_secret_post, client_secret_basic, client_secret_jwt, and
+     * private_key_jwt, as described in Section 9 of OpenID Connect Core 1.0 [OpenID.Core]. Other
+     * authentication methods MAY be defined by extensions. If omitted, the default is client_secret_basic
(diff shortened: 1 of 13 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/oidcdiscovery/OIDCDiscoveryEndpoint.java
```diff
@@ -98,7 +99,7 @@ private Response getResponse(HttpServletRequest request, String tenant) {
         String response;
         OIDCProcessor processor = EndpointUtil.getOIDCService();
         try {
-            OIDProviderResponseBuilder responseBuilder = new OIDProviderJSONResponseBuilder();
+            OIDProviderResponseBuilder responseBuilder = getOidProviderResponseBuilder();
```
mefarazath: Can we consider renaming the class name OIDProviderResponseBuilder?
OIDCProviderDiscoveryResponseBuilder or OpenIDConnectProviderDiscoveryResponseBuilder 

or something along the lines that would give the purpose of the extension or interface.

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/DiscoveryConstants.java
```diff
@@ -47,6 +47,12 @@
      * MAY contain port, path, and query parameter components.
      */
     public static final String USERINFO_ENDPOINT = "Userinfo_endpoint";
+    /**
```
mefarazath: Can we do a comparison with the spec and see if there are other missing ones we can support OOTB?

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/oidcdiscovery/OIDCDiscoveryEndpoint.java
emswbandara: For backward compatibility we will keep the name as it is. We will improve the Java Docs of this interface to convey the actual purpose of the interface.

emswbandara:
I have analyzed the missing optional configurations that we can introduce to our OIDC discovery endpoint response and those are as follows.
```
response_modes_supported
token_endpoint_auth_signing_alg_values_supported
service_documentation
ui_locales_supported
op_policy_uri
revocation_endpoint_auth_methods_supported
revocation_endpoint_auth_signing_alg_values_supported
introspection_endpoint
introspection_endpoint_auth_methods_supported
introspection_endpoint_auth_signing_alg_values_supported
code_challenge_methods_supported
```

Out of the above, with this PR i have provided support for the following elements. 
```
response_modes_supported
revocation_endpoint
revocation_endpoint_auth_methods_supported
introspection_endpoint
introspection_endpoint_auth_methods_supported
code_challenge_methods_supported
```
However as per the current implementation, most of the values of these elements are hard-coded and the capability to configure those will be addressed via [1] since it does not belong to the scope of fixing this issue.

```
token_endpoint_auth_signing_alg_values_supported
revocation_endpoint_auth_signing_alg_values_supported
introspection_endpoint_auth_signing_alg_values_supported
```

For all of the above fields the the spec suggests that it should return JSON array containing a list of the JWS signing algorithms ("alg" values) supported by the token endpoint for the signature on the JWT [JWT] used to authenticate the client at the token endpoint for the "private_key_jwt" and "client_secret_jwt" authentication methods.  This metadata entry MUST be present if either of these authentication methods are specified in the "token_endpoint_auth_methods_supported" entry. 
As we support configuring private-key-jwt via an extension I believe we cannot send the response for this field by default in the product. Hence it needs to be addressed via [1]

**service_documentation**
This indicates URL of a page containing human-readable information that developers might want or need to know when using the authorization server.  In particular, if the authorization server does not support Dynamic Client Registration, then information on how to register clients needs to be provided in this documentation.
As we support DCR, there is no requirement for this field from discovery response.

**op_policy_uri** : URL that the authorization server provides to the person registering the client to read about the authorization server's requirements on how the client can use the data provided
      by the authorization server 
**op_tos_uri** : URL that the authorization server provides to the person registering the client to read about the authorization server's terms of service.

Above two will be addressed via [1]

**ui_locales_supported**
This indicates the Languages and scripts supported for the user interface. Since we support localization feature for oauth2.war IMO it is better to provide the capability to configure the value of this field rather that providing hard-coded value. This will also be addressed via [1]

[1] https://github.com/wso2/product-is/issues/4381

@mefarazath please let me know if you have any concerns.

PR #1331, opened 2020-02-24
Fix allowed scopes retrival of non admin users

IsuraD:

Fix allowed scopes retrival of non admin users


IsuraD:
Fix https://github.com/wso2/product-is/issues/7741

PR #1345, opened 2020-03-23
Tenant qualify OIDC discovery endpoint response

ChamathNS:
#### Proposed changes in this pull request

As per the requirement to make all the endpoints tenant-qualified, the OIDC discovery endpoint should support the form 'https://localhost:9443/t/{tenant.domain}/oauth2/token/.well-known/openid-configuration'

Currently, the endpoint supports tenants in path. But, the URLs that are generated are not tenant-qualified.
Eg: https://localhost:9443/oauth2/token

With the changes in the PR, the OIDC metadata URLs would be tenant-qualified with tenant in path.
Eg: https://localhost:9443/t/wso2.com/oauth2/token

fix https://github.com/wso2/product-is/issues/7932

### DIFF
--- a/components/org.wso2.carbon.identity.discovery/src/test/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilderTest.java
+++ b/components/org.wso2.carbon.identity.discovery/src/test/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilderTest.java
@@ -29,6 +29,7 @@ import org.wso2.carbon.base.ServerConfigurationException;
 import org.wso2.carbon.identity.claim.metadata.mgt.ClaimMetadataManagementService;
 import org.wso2.carbon.identity.claim.metadata.mgt.exception.ClaimMetadataException;
 import org.wso2.carbon.identity.claim.metadata.mgt.model.ExternalClaim;
+import org.wso2.carbon.identity.core.util.IdentityUtil;
 import org.wso2.carbon.identity.discovery.OIDProviderRequest;
 import org.wso2.carbon.identity.discovery.internal.OIDCDiscoveryDataHolder;
 import org.wso2.carbon.identity.oauth.config.OAuthServerConfiguration;
@@ -39,6 +40,7 @@ import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.List;
 
+import static org.mockito.Matchers.anyBoolean;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.MockitoAnnotations.initMocks;
 import static org.powermock.api.mockito.PowerMockito.mock;
@@ -48,7 +50,7 @@ import static org.powermock.api.mockito.PowerMockito.when;
 import static org.testng.Assert.assertNotNull;
 
 @PrepareForTest({OAuth2Util.class, OAuthServerConfiguration.class, OIDCDiscoveryDataHolder.class,
-        ClaimMetadataManagementService.class})
+        ClaimMetadataManagementService.class, IdentityUtil.class})
 /**
  * Unit test covering ProviderConfigBuilder class.
  */
@@ -90,6 +92,7 @@ public class ProviderConfigBuilderTest {
 
         mockStatic(OAuth2Util.class);
         mockStatic(OAuth2Util.OAuthURL.class);
+        mockStatic(IdentityUtil.class);
 
         List<ExternalClaim> claims = new ArrayList<>();
         ExternalClaim externalClaim = new ExternalClaim("aaa", "bbb", "ccc");
@@ -101,6 +104,7 @@ public class ProviderConfigBuilderTest {
 
         when(OAuth2Util.mapSignatureAlgorithmForJWSAlgorithm(idTokenSignatureAlgorithm)).thenReturn(JWSAlgorithm.RS256);
         when(OAuth2Util.mapSignatureAlgorithmForJWSAlgorithm(anyString())).thenReturn(JWSAlgorithm.RS256);
+        when(IdentityUtil.resolveURL(anyString(), anyBoolean(), anyBoolean())).thenReturn("/testUrl");
         assertNotNull(providerConfigBuilder.buildOIDProviderConfig(mockOidProviderRequest));
     }
 
@@ -113,7 +117,7 @@ public class ProviderConfigBuilderTest {
 
         mockStatic(OAuth2Util.class);
         mockStatic(OAuth2Util.OAuthURL.class);
-        when(OAuth2Util.OAuthURL.getOAuth2JWKSPageUrl(anyString())).thenThrow(new URISyntaxException("input",
+        when(OAuth2Util.OAuthURL.getOAuth2JWKSPageUrl()).thenThrow(new URISyntaxException("input",
                 "URISyntaxException"));
         when(OAuth2Util.getIdTokenIssuer(anyString())).thenReturn("issuer");
         providerConfigBuilder.buildOIDProviderConfig(mockOidProviderRequest);
--- a/components/org.wso2.carbon.identity.discovery/src/test/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilderTest.java
+++ b/components/org.wso2.carbon.identity.discovery/src/test/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilderTest.java
@@ -109,21 +109,6 @@ public class ProviderConfigBuilderTest {
     }
 
     @Test(expectedExceptions = ServerConfigurationException.class)
-    public void testBuildOIDProviderConfig1Legacy() throws Exception {
-
-        OAuthServerConfiguration mockOAuthServerConfiguration = mock(OAuthServerConfiguration.class);
-        mockStatic(OAuthServerConfiguration.class);
-        when(OAuthServerConfiguration.getInstance()).thenReturn(mockOAuthServerConfiguration);
-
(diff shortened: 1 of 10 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1168,33 +1168,25 @@ public static String getOAuth2TokenEPUrl() {
             if (StringUtils.isBlank(oauth2TokenEPUrl)) {
                 oauth2TokenEPUrl = IdentityUtil.getServerURL("oauth2/token", true, false);
             }
-            return oauth2TokenEPUrl;
+            return IdentityUtil.resolveURL(oauth2TokenEPUrl, true, false);
         }
 
-        public static String getOAuth2DCREPUrl(String tenantDomain) throws URISyntaxException {
```
mefarazath: Changing the public method all of a sudden is not good practice as it might break any code that is relying on this particular Util class. 

Best way to introduce this change would be to introduce a new method, deprecate the old method and point to the newly introduced method for future usage.

Then, in a major version of the component, we can remove the deprecated method.

mefarazath: Also we can change the old method's implementation to use the newly introduced method

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -3086,8 +3078,8 @@ public static String getIdTokenIssuer(String tenantDomain) throws IdentityOAuth2
         FederatedAuthenticatorConfig oidcAuthenticatorConfig =
                 IdentityApplicationManagementUtil.getFederatedAuthenticator(fedAuthnConfigs,
                         IdentityApplicationConstants.Authenticator.OIDC.NAME);
-        return IdentityApplicationManagementUtil.getProperty(oidcAuthenticatorConfig.getProperties(),
-                IDP_ENTITY_ID).getValue();
+        return IdentityUtil.resolveURL(IdentityApplicationManagementUtil
```
mefarazath: I think this is not needed. What we do here is to return the value set at the resident IDP. It can be a random URI value. Trying to tenant qualify that is incorrect.

mefarazath: For example, I set a value like "https://myissuer.xyz.com" at the resident IDP as the issuer value. When we use the IdentityUtil.resolveURL() method what would be the output?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
ChamathNS: > Also we can change the old method's implementation to use the newly introduced method

Agreed on introducing a new method.
With the URL resolver service, DefaultURLResolverService.class takes the tenant name from the ThreadLocal when qualifying paths for tenants. Therefore, the new method would not require the tenant domain as a parameter. Hence, wouldn't it be incorrect if we change the old method to consume the new method?

mefarazath: Yes. Let's leave the old method as it is and a comment to use the new method.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
ChamathNS: fixed with https://github.com/wso2/carbon-identity-framework/pull/2836/commits/f8b4a59dc81f0b5f1d29aa06e3109d6b80721781

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1184,6 +1184,15 @@ public static String getOAuth2DCREPUrl(String tenantDomain) throws URISyntaxExce
             return oauth2TokenEPUrl;
         }
 
+        public static String getOAuth2DCREPUrl() throws URISyntaxException {
+
+            String oauth2TokenEPUrl = OAuthServerConfiguration.getInstance().getOAuth2DCREPUrl();
+            if (StringUtils.isBlank(oauth2TokenEPUrl)) {
+                oauth2TokenEPUrl = IdentityUtil.getServerURL("/api/identity/oauth2/dcr/v1.0/register", true, false);
```
mefarazath: It would be better to define a constant for "/api/identity/oauth2/dcr/v1.0/register"

ChamathNS: This is fixed after adding String constants for OAuth2 endpoints.

pom.xml
```diff
@@ -865,7 +865,7 @@
         <carbon.kernel.registry.imp.pkg.version.range>[1.0.1, 2.0.0)</carbon.kernel.registry.imp.pkg.version.range>
 
         <!-- Carbon Identity Framework version -->
-        <carbon.identity.framework.version>5.16.119</carbon.identity.framework.version>
+        <carbon.identity.framework.version>5.17.10-SNAPSHOT</carbon.identity.framework.version>
```
emswbandara: remove snapshot version and update to a released version

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -3119,8 +3119,13 @@ public static String getIdTokenIssuer(String tenantDomain) throws IdentityOAuth2
         FederatedAuthenticatorConfig oidcAuthenticatorConfig =
                 IdentityApplicationManagementUtil.getFederatedAuthenticator(fedAuthnConfigs,
                         IdentityApplicationConstants.Authenticator.OIDC.NAME);
-        return IdentityApplicationManagementUtil.getProperty(oidcAuthenticatorConfig.getProperties(),
+        String idpEntityId = IdentityApplicationManagementUtil.getProperty(oidcAuthenticatorConfig.getProperties(),
                 IDP_ENTITY_ID).getValue();
+        if (idpEntityId.equals(IdentityUtil.getProperty("OAuth.OpenIDConnect.IDTokenIssuerID"))) {
```
mefarazath: This will lead to NPE if idpEntityId evaluates to null. Can we be sure that it can never be null?
Also can IdentityUtil.getProperty("OAuth.OpenIDConnect.IDTokenIssuerID") evaluate to null in any scenario?

components/org.wso2.carbon.identity.discovery/src/test/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilderTest.java
```diff
@@ -113,7 +117,7 @@ public void testBuildOIDProviderConfig1() throws Exception {
 
         mockStatic(OAuth2Util.class);
         mockStatic(OAuth2Util.OAuthURL.class);
-        when(OAuth2Util.OAuthURL.getOAuth2JWKSPageUrl(anyString())).thenThrow(new URISyntaxException("input",
+        when(OAuth2Util.OAuthURL.getOAuth2JWKSPageUrl()).thenThrow(new URISyntaxException("input",
                 "URISyntaxException"));
         when(OAuth2Util.getIdTokenIssuer(anyString())).thenReturn("issuer");
         providerConfigBuilder.buildOIDProviderConfig(mockOidProviderRequest);
```
malithie: We should add test cases for legacy mode and new mode and verify OIDC endpoints.
Shall we do that as well.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1197,20 +1221,29 @@ public static String getOAuth2JWKSPageUrl(String tenantDomain) throws URISyntaxE
             return auth2JWKSPageUrl;
         }
 
+        public static String getOAuth2JWKSPageUrl() throws URISyntaxException {
```
dewniMW: Shall we add a method level comment?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1184,11 +1199,20 @@ public static String getOAuth2DCREPUrl(String tenantDomain) throws URISyntaxExce
             return oauth2TokenEPUrl;
         }
 
+        public static String getOAuth2DCREPUrl() throws URISyntaxException {
```
dewniMW: Shall we add a method level comment?

PR #1347, opened 2020-03-31
OAuth Consent page URL generation set to comply with tenant qualification

ashendes:
**Related issues:**
[product-is/issues/7993](https://github.com/wso2/product-is/issues/7993)

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
@@ -2483,7 +2483,8 @@ public class OAuth2AuthzEndpoint {
                 if (attribute == AuthenticatorFlowStatus.INCOMPLETE) {
 
                     if (responseWrapper.isRedirect()) {
-                        oAuthMessage.getResponse().sendRedirect(responseWrapper.getRedirectURL());
+                        return Response.status(HttpServletResponse.SC_FOUND).location(
+                                new URI(responseWrapper.getRedirectURL())).build();
                     } else {
                         return Response.status(HttpServletResponse.SC_OK).entity(responseWrapper.getContent()).build();
                     }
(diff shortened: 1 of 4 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/util/EndpointUtil.java
```diff
@@ -610,9 +610,9 @@ public static String getUserConsentURL(OAuth2Parameters params, String loggedInU
             }
 
             if (isOIDC) {
-                consentPage = OAuth2Util.OAuthURL.getOIDCConsentPageUrl();
+                consentPage = IdentityUtil.resolveURL(OAuth2Util.OAuthURL.getOIDCConsentPageUrl(), true, true);
```
mefarazath: Should we do this?
With changes done by @ChamathNS, OAuth2Util.OAuthURL.getOIDCConsentPageUrl() should return the tenant qualified URLs, isn't it?

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
```diff
@@ -2483,7 +2483,8 @@ private Response handleAuthFlowThroughFramework(OAuthMessage oAuthMessage, Strin
                 if (attribute == AuthenticatorFlowStatus.INCOMPLETE) {
 
                     if (responseWrapper.isRedirect()) {
-                        oAuthMessage.getResponse().sendRedirect(responseWrapper.getRedirectURL());
+                        return Response.status(HttpServletResponse.SC_FOUND).location(
```
mefarazath: What's the reason for this change?

malithie: @ashendes , did we check removing this.
The first explation was this was added as this "returns 204 instead of 302 when servlet forwards are done at the rest dispatcher"

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/util/EndpointUtil.java
malithie: This should be fixed in OAuth2Util.OAuthURL such that URL is resolved always an tenant url supported mode, until @ChamathNS get that fixed from the API

PR #1353, opened 2020-04-15
Fix OIDC Discovery metadata endpoint

ChamathNS:
### Proposed changes in this pull request

Change the IdentityUtil.resolveURL instances to use ServiceURLBuilder API

Improvement for #1345 
Fix https://github.com/wso2/product-is/issues/7932

PR #1361, opened 2020-04-22
Remove redundant methods that use resolveURL

ChamathNS:

remove redundant getOAuth2JWKSPageUrl() method which consumed , now deprecated and removed, IdentityUtil.resolveURL API

ASAP

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
@@ -1220,21 +1220,6 @@ public class OAuth2Util {
             return auth2JWKSPageUrl;
         }
 
-        /**
-         * This method is used to get the resolved URL for the JWKS Page.
-         *
-         * @return String of the resolved URL for the JWKS page.
-         * @throws URISyntaxException URI Syntax Exception.
-         */
-        public static String getOAuth2JWKSPageUrl() throws URISyntaxException {
-
-            String auth2JWKSPageUrl = OAuthServerConfiguration.getInstance().getOAuth2JWKSPageUrl();
-            if (StringUtils.isBlank(auth2JWKSPageUrl)) {
-                auth2JWKSPageUrl = IdentityUtil.getServerURL(OAUTH2_JWKS_EP_URL, true, false);
-            }
-            return IdentityUtil.resolveURL(auth2JWKSPageUrl, true, true, false, true);
-        }
-
         public static String getOidcWebFingerEPUrl() {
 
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1220,21 +1220,6 @@ public static String getOAuth2JWKSPageUrl(String tenantDomain) throws URISyntaxE
             return auth2JWKSPageUrl;
         }
 
-        /**
-         * This method is used to get the resolved URL for the JWKS Page.
-         *
-         * @return String of the resolved URL for the JWKS page.
-         * @throws URISyntaxException URI Syntax Exception.
-         */
-        public static String getOAuth2JWKSPageUrl() throws URISyntaxException {
```
janakamarasena, line 1229: Since this is a public method wondering whether we should first deprecate this and then remove it in the future.

mefarazath: Yes, +1.
Let's deprecate and remove with a version bump

ChamathNS: This method was introduced as an overloaded method for getOAuth2JWKSPageUrl(String tenantDomain) [1] which is being used widely.
The purpose of introducing getOAuth2JWKSPageUrl() with [2]  was to consume the IdentityUtil.resolveURL API which has been now removed with [3].
Since the usages of this method are now set to consume the old [1] method with [4], I thought it would be safe to remove this method. Since we may need to introduce framework changes with the feature, "tenant-qualifying endpoints", we would need to remove this method soon since we wouldn't be able to bump the framework version in this module to incorporate those changes if this method is kept.


[1] - https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/92711d34215089fd6cf0eb376f9f39a0fec77cac/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java#L1211

[2] - https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1345

[3] - https://github.com/wso2/carbon-identity-framework/pull/2865

[4] - https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1353

janakamarasena: But there be any outside extensions that could be depending on this. So IMO deprecating 1st would be best. 
@mefarazath WDYT?

PR #1364, opened 2020-04-24
Role based scope validator implementation for WSO2 IS 

chamilaadhi:

This PR is done for the feature requested in https://github.com/wso2/product-is/issues/8163

The behavior of this validator is as below
- It only works if it is specifically enabled.
- The scope is only validated if the requested scope is registered with bind type 'DEFAULT'. If the scope is registered, then roles related to the registered scope is used to validate the requested scope and only issue if the user has the same role
- If the scope is not registered with bind type 'DEFAULT', send the token back with the requested scope (to facilitate IS default behavior)

This implementation is done by moving the WSO2 APIM Scope Issuer [RoleBasedScopesIssuer.java]( https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/issuers/RoleBasedScopesIssuer.java#L99) code to a validator class in IS. Also to work with SAML and JWT bearer grant type, I have included the scope validation related code used in [ExtendedSAML2BearerGrantHandler.java](https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/handlers/ExtendedSAML2BearerGrantHandler.java#L32) and [ExtendedJWTBearerGrantHandler.java](https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.keymgt/src/main/java/org/wso2/carbon/apimgt/keymgt/token/ExtendedJWTBearerGrantHandler.java#L61)

To enable this , following needs to be enabled in the identity.xml file.
` <EnableRoleBasedScopeValidator>false</EnableRoleBasedScopeValidator>`

Above config is mapped to `oauth.enable_role_based_scope_validator` in the `deployment.toml` file.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
@@ -190,6 +190,8 @@ public class OAuthServerConfiguration {
     private boolean useMultiValueSeparatorForAuthContextToken = true;
     private boolean addTenantDomainToIdTokenEnabled = false;
     private boolean addUserstoreDomainToIdTokenEnabled = false;
+    
+    private boolean isRoleBasedScopeValidatorEnabled = false;
 
     //default token types
     public static final String DEFAULT_TOKEN_TYPE = "Default";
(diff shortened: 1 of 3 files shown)


chamilaadhi:
closing this PR and will send a PR again with the changes requested in https://github.com/wso2/product-is/issues/8245

PR #1370, opened 2020-04-29
Validate request tenant domain in tenant qualified URL mode

mefarazath:
In tenant qualified URL mode we have the tenant domain sent in path param as a value in the thread-local context. 

For example https://localhost:9443/t/wso2.com/oauth2/token

- With this PR, we validate the tenant domain in the context against the tenant domain derived from the client_id. This will prevent cross tenant access.

For example, if you create an OAuth app in tenant domain foo.com and use the URL of bar.com as the token endpoint https://localhost:9443/t/bar.com/oauth2/token for the request, you will get an error

```
{
    "error_description": "Invalid Client",
    "error": "invalid_client"
}
```

- Pass error message set in the AccessTokenIssuer level to the Response build by the endpoint.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpoint.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpoint.java
@@ -19,12 +19,12 @@
 package org.wso2.carbon.identity.oauth.endpoint.token;
 
 import org.apache.commons.collections.MapUtils;
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.cxf.interceptor.InInterceptors;
 import org.apache.oltu.oauth2.as.response.OAuthASResponse;
 import org.apache.oltu.oauth2.as.response.OAuthASResponse.OAuthTokenResponseBuilder;
-import org.apache.oltu.oauth2.common.OAuth;
 import org.apache.oltu.oauth2.common.error.OAuthError;
 import org.apache.oltu.oauth2.common.exception.OAuthProblemException;
 import org.apache.oltu.oauth2.common.exception.OAuthSystemException;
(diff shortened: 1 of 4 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpoint.java
```diff
@@ -328,7 +276,7 @@ private OAuth2AccessTokenReqDTO buildAccessTokenReqDTO(CarbonOAuthTokenRequest o
         tokenReqDTO.setClientId(oauthClientAuthnContext.getClientId());
         tokenReqDTO.setClientSecret(oauthRequest.getClientSecret());
         tokenReqDTO.setCallbackURI(oauthRequest.getRedirectURI());
-        tokenReqDTO.setScope(oauthRequest.getScopes().toArray(new String[oauthRequest.getScopes().size()]));
+        tokenReqDTO.setScope(oauthRequest.getScopes().toArray(new String[0]));
```
denuwanthi, line 279: why are we creating array with one element here?

mefarazath: It was a suggestion from the IDE.

The inspection says

> Call to 'toArray()' with pre-sized array argument 'new IdentityProviderProperty[newProperties.size()]' 

>  Inspection info: There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).

> In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
> This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

PR #1386, opened 2020-05-27
[Fix] User claims not retuning in an access token for a federated user via implicit grant type 

somindatommy:
Resolves https://github.com/wso2/product-is/issues/8377

When obtaining the access token, the user claims are not cached.
In that scenario, check whether the authenticated user is federated or not. If federated, get the user claims from `OAuthAuthzReqMessageContext`.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultOIDCClaimsCallbackHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultOIDCClaimsCallbackHandler.java
@@ -127,7 +127,7 @@ public class DefaultOIDCClaimsCallbackHandler implements CustomClaimsCallbackHan
     }
 
     /**
-     * Get response map
+     * Get response map.
      *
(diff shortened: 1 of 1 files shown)


mounirloubari:
Hello @somindatommy @tharindu-bandara 

I did this change but i still have not my custom claim in issued jwt token.
I made sur that:

1. Create a custom local claim ( http://wso2.org/claims/customclaim ) 
2. Create an external claim (customclaim) and mapped it to local custom claim.
3. In IDP level i set the mapping beteween my requested claim and local custom claim ( ex: "sub" mapped to "http://wso2.org/claims/customclaim")
4. In SP level  i set the local custom claim  in claim configuration.
5. I made sur also that my external custom claim (customclaim) is in openid scope.

When i request a jwt token with implicite grant type i have no trace of my custom claim.

**Knowing:**
I'am using IS 5.10.0 with identity-inbound-auth-oauth extension v6.4.2.

I have this debug log: 
claimUrisInRequestedScope customclaim
DEBUG {org.wso2.carbon.identity.openidconnect.OpenIDConnectClaimFilterImpl} - No valid user claim value found for the claimUri:customclaim

I  you have any idea i will appreciate it.

PR #1395, opened 2020-06-12
Remove session_state from error flows

mefarazath:


mefarazath:
Fix has been added with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1400

Issue #1398, opened 2020-06-17
session_state parameter is appended to the error URL when authentication is failed


Akila94:
**Description:**
According to [OIDC specification](https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.3.1.2.6), the server should not send any parameters except error code and state if authentication fails Unless the Redirection URI is invalid. In the current implementation, the "[session_state](https://openid.net/specs/openid-connect-session-1_0.html#CreatingUpdatingSessions)" parameter is appended to the error URL. As the authentication is failed in the scenario, there is no point of appending the "session_state" parameter to the error URL as upon the authorization failure, the user is not logged in and there is no need for the  RP to get this session_state parameter when there is no session.

Appending the "session_state" parameter to the error URL when the authentication failed should be avoided.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**
IS 5.7

**OS, DB, other environment details and versions:**
MacOS, MySQL, Java 8    

**Steps to reproduce:**
1. Demonstrate a scenario in which the authentication is failed using IS 5.7.
2. We can see the "session_state" is appended to the error URL.

**Related Issues:**

PR #1400, opened 2020-06-19
Remove appending session_state parameter for auth failure error URL

Akila94:
Fix for the issue: wso2-extensions/identity-inbound-auth-oauth#1398


Akila94:
Hi @janakamarasena , I tested the fix in the master pack and verified that the fix works as expected.

PR #1410, opened 2020-07-09
Fix tenant flow issue

ivantha:
Fixes: https://github.com/wso2-extensions/identity-inbound-auth-oauth/issues/1409

Related to: https://github.com/wso2-extensions/identity-carbon-auth-rest/pull/114

### DIFF
--- a/components/org.wso2.carbon.identity.discovery/pom.xml
+++ b/components/org.wso2.carbon.identity.discovery/pom.xml
@@ -21,7 +21,7 @@
         <groupId>org.wso2.carbon.identity.inbound.auth.oauth2</groupId>
         <artifactId>identity-inbound-auth-oauth</artifactId>
         <relativePath>../../pom.xml</relativePath>
-        <version>6.4.49</version>
+        <version>6.4.50-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
--- a/components/org.wso2.carbon.identity.oauth.ciba/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth.ciba/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <artifactId>identity-inbound-auth-oauth</artifactId>
         <groupId>org.wso2.carbon.identity.inbound.auth.oauth2</groupId>
-        <version>6.4.49</version>
+        <version>6.4.50-SNAPSHOT</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
--- a/components/org.wso2.carbon.identity.oauth.client.authn.filter/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth.client.authn.filter/pom.xml
@@ -22,7 +22,7 @@
         <groupId>org.wso2.carbon.identity.inbound.auth.oauth2</groupId>
         <artifactId>identity-inbound-auth-oauth</artifactId>
         <relativePath>../../pom.xml</relativePath>
-        <version>6.4.49</version>
+        <version>6.4.50-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
--- a/components/org.wso2.carbon.identity.oauth.common/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth.common/pom.xml
@@ -23,7 +23,7 @@
         <groupId>org.wso2.carbon.identity.inbound.auth.oauth2</groupId>
         <artifactId>identity-inbound-auth-oauth</artifactId>
         <relativePath>../../pom.xml</relativePath>
-        <version>6.4.49</version>
+        <version>6.4.50-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
--- a/components/org.wso2.carbon.identity.oauth.dcr.endpoint/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth.dcr.endpoint/pom.xml
@@ -6,7 +6,7 @@
         <groupId>org.wso2.carbon.identity.inbound.auth.oauth2</groupId>
         <artifactId>identity-inbound-auth-oauth</artifactId>
         <relativePath>../../pom.xml</relativePath>
-        <version>6.4.49</version>
+        <version>6.4.50-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
--- a/components/org.wso2.carbon.identity.oauth.dcr/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth.dcr/pom.xml
@@ -22,7 +22,7 @@
         <groupId>org.wso2.carbon.identity.inbound.auth.oauth2</groupId>
         <artifactId>identity-inbound-auth-oauth</artifactId>
         <relativePath>../../pom.xml</relativePath>
-        <version>6.4.49</version>
+        <version>6.4.50-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
--- a/components/org.wso2.carbon.identity.oauth.endpoint/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/pom.xml
@@ -22,7 +22,7 @@
         <groupId>org.wso2.carbon.identity.inbound.auth.oauth2</groupId>
         <artifactId>identity-inbound-auth-oauth</artifactId>
         <relativePath>../../pom.xml</relativePath>
-        <version>6.4.49</version>
+        <version>6.4.50-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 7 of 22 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/jwks/JwksEndpoint.java
```diff
@@ -103,7 +103,9 @@ public String jwks() {
             String errorMessage = "Error while generating the keyset for tenant domain: " + tenantDomain;
             return logAndReturnError(errorMessage, e);
         } finally {
-            FrameworkUtils.endTenantFlow();
+            if (!MultitenantConstants.SUPER_TENANT_DOMAIN_NAME.equalsIgnoreCase(tenantDomain)) {
```
mefarazath, line 106: This fix this perfectly fine.

I was wondering whether we could get rid of the if block by,

#1 - Move the validating tenant ID logic outside the try block.
#2 - Start the tenant flow at the beginning of the try block (regardless of whatever the tenant it is)
#3 - end tenant flow at the finally block

That would make this code block readable and less error-prone. WDYT?

mefarazath: Also even, after the proposed fix we can face the same issue if 
`IdentityTenantUtil.getTenantId(tenantDomain)` line throws an exception which would result in finally block to attempt to end a tenant flow that was never started.

ivantha: @mefarazath That occurred to me. It looks better; however, I was wondering about the possibility of it breaking something since the original code had put it specifically inside the block for non-super tenants.

mefarazath: If it works for super tenant and tenant endpoints we are good.

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/jwks/JwksEndpoint.java
```diff
@@ -73,20 +73,22 @@ public String jwks() {
         String tenantDomain = getTenantDomain();
         String keystorePath = CarbonUtils.getServerConfiguration().getFirstProperty(SECURITY_KEY_STORE_LOCATION);
 
+        int tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
+        if (isInvalidTenantId(tenantId)) {
```
emswbandara, line 77: do we need to validate the tenant here since we are retrieving the tenant domain from thread local context?

mefarazath: Yes. But if someone calls the jwks endpoint of a non-existing tenant domain still it would set in the thread-local and passed here. isn't it?

emswbandara: it seems we are already doing the validation in rewrite valve.. https://github.com/wso2-extensions/identity-carbon-auth-rest/blob/3d37c882f4b96701b8d333c2fc82b42b18da0348/components/org.wso2.carbon.identity.context.rewrite.valve/src/main/java/org/wso2/carbon/identity/context/rewrite/valve/TenantContextRewriteValve.java#L111-L112
I think we can remove the redundant validation here. WDYT

mefarazath: +1

ivantha: :+1:

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/jwks/JwksEndpoint.java
```diff
@@ -72,21 +72,18 @@ public String jwks() {
 
         String tenantDomain = getTenantDomain();
         String keystorePath = CarbonUtils.getServerConfiguration().getFirstProperty(SECURITY_KEY_STORE_LOCATION);
```
mefarazath, line 74: Another small thing I noticed was that, 

Loading the keystore form `keystorePath` is only required for super.tenant.

But we are opening a file input stream regardless of whether its super tenant or not

ivantha: Fixed :)

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/jwks/JwksEndpoint.java
```diff
@@ -71,22 +71,21 @@
...
-        try (FileInputStream file = new FileInputStream(keystorePath)) {
-            int tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
             final KeyStore keystore;
             Map<String, Certificate> certificatesWithAliases = new HashMap<>();
             if (MultitenantConstants.SUPER_TENANT_DOMAIN_NAME.equalsIgnoreCase(tenantDomain)) {
+                String keystorePath = CarbonUtils.getServerConfiguration()
+                        .getFirstProperty(SECURITY_KEY_STORE_LOCATION);
                 keystore = KeyStore.getInstance(KeyStore.getDefaultType());
                 String password = CarbonUtils.getServerConfiguration().getFirstProperty(SECURITY_KEY_STORE_PW);
+                FileInputStream file = new FileInputStream(keystorePath);
```
mefarazath, line 86: Better to do this within a try with resource block to make sure FileInputStream closed

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/jwks/JwksEndpoint.java
ivantha: If we move the `FileInputStream` away from the try-with pattern and into the "if" block of the super tenant, we will have to close the resource inside the try itself. Is this alright?

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/jwks/JwksEndpoint.java
ivantha: With another `try` inside the main try block?

ivantha: Or are you suggesting breaking down the try block into several separate try-catch blocks?

emswbandara: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/jwks/JwksEndpoint.java
```diff
@@ -71,22 +71,21 @@
...
             if (MultitenantConstants.SUPER_TENANT_DOMAIN_NAME.equalsIgnoreCase(tenantDomain)) {
+                String keystorePath = CarbonUtils.getServerConfiguration().getFirstProperty(SECURITY_KEY_STORE_LOCATION);
                 keystore = KeyStore.getInstance(KeyStore.getDefaultType());
                 String password = CarbonUtils.getServerConfiguration().getFirstProperty(SECURITY_KEY_STORE_PW);
-                keystore.load(file, password.toCharArray());
-            } else {
-                if (isInvalidTenantId(tenantId)) {
-                    String errorMessage = "Invalid Tenant: " + tenantDomain;
-                    return logAndReturnError(errorMessage, null);
+                try(FileInputStream file = new FileInputStream(keystorePath)){
```
emswbandara, line 85: formatting issue

emswbandara:
@ivantha can you please check why travis is failing

PR #1416, opened 2020-07-24
Delete SP apps on tenant deletion

sumedhe:

Calls SP application deletion on tenant deletion

Related: wso2/carbon-identity-framework#2932


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/listener/TenantCreationEventListener.java
```diff
@@ -101,6 +102,12 @@ public void onPreDelete(int tenantId) throws StratosException {
         } catch (IdentityOAuthAdminException e) {
             throw new StratosException("Error in deleting all OAuth application data of the tenant: " + tenantId, e);
         }
+
```
IsuraD, line 105: This should be done in framework level? Didn't we delete the applications in a framework listener?

IsuraD:
This is not needed with the new tenant deletion approach

PR #1418, opened 2020-07-30
Fix cluster message doesn't removes the OAuth token

Sathiyakugan:
Resolves: https://github.com/wso2/product-is/issues/9053

**This PR fixes the not matching cache entry when removing.** 

Fix adds the Domain Name to the username when removing the cache to match with the cache entry key inserted.


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/OAuthCacheRemoveListener.java
```diff
@@ -53,18 +53,18 @@ public void entryRemoved(CacheEntryEvent<? extends OAuthCacheKey, ? extends Cach
             log.debug("OAuth cache removed for consumer id : " + accessTokenDO.getConsumerKey());
         }
 
+        String userName = accessTokenDO.getAuthzUser().toString();
         boolean isUsernameCaseSensitive = IdentityUtil
                 .isUserStoreInUsernameCaseSensitive(accessTokenDO.getAuthzUser().getUserName());
```
nilasini, line 58: Please send the username retrieved from  accessTokenDO.getAuthzUser().toString() to  isUserStoreInUsernameCaseSensitive method.

PR #1454, opened 2020-09-22
Fix refresh token grant issue for auth-code grant type token with a request object

nalakagunathilaka:

Fixes https://github.com/wso2/product-is/issues/9525.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
@@ -1648,7 +1648,7 @@ public class AccessTokenDAOImpl extends AbstractOAuthDAO implements AccessTokenD
             }
         }
 
-        Connection connection = IdentityDatabaseUtil.getDBConnection();
+        Connection connection = IdentityDatabaseUtil.getDBConnection(true);
         try {
             // update existing token as inactive
             updateAccessTokenState(connection, oldAccessTokenId, tokenState, tokenStateId, userStoreDomain);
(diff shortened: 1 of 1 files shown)


ruwanta (dismissed):
We need             IdentityDatabaseUtil.rollbackTransaction(connection); at the exception handling phase. Otherwise some DBs (e.g. PG SQL) will have too many open connections.

ruwanta (approved):

ruwanta (changes requested):
Seems like IdentityDatabaseUtil.commitTransaction(connection); already present

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
```diff
@@ -1660,6 +1660,8 @@ public void invalidateAndCreateNewAccessToken(String oldAccessTokenId, String to
             // update new access token against authorization code if token obtained via authorization code grant type
             updateTokenIdIfAutzCodeGrantType(oldAccessTokenId, accessTokenDO.getTokenId(), connection);
 
+            IdentityDatabaseUtil.commitTransaction(connection);
```
ruwanta, line 1663: Seems like IdentityDatabaseUtil.commitTransaction(connection); already present at line 1671. This needs to be removed, otherwise there will be exceptions saying about transaction already commited.

nalakagunathilaka: We need to commit the transaction here since postRefreshAccessToken() method needs the committed data from database. This commit was there previously before the code refactoring took place. Once we remove this commit operation here, refresh token process doesn't work for auth code grant type access token using the request object as we tested.

ruwanta: There can be only one commit transaction for one begin transaction.
Also there can not be multiple transactions in the single flow.

Above comment means there are some more improvement to be done.

nalakagunathilaka: Thank you very much for the insights, will do the needful ASAP.

thanujalk:
Closing as per the conversation.

PR #1447, opened 2020-09-15
Deprecate OAuth 1.0 and /identity/connect/register APIs

GANGANI:
### Proposed changes in this pull request

Deprecate OAuth 1.0 and /identity/connect/register APIs 

**Related issues**: https://github.com/wso2/product-is/issues/8564, https://github.com/wso2/product-is/issues/8614

**Related PR**: https://github.com/wso2/carbon-identity-framework/pull/3027

PR #1460, opened 2020-09-29
Remove invalid deprecation

GANGANI:

Remove invalid deprecations.


janakamarasena:
Why were these deprecations added in the 1st place and why are they being removed now?

emswbandara (approved):

GANGANI:
> Why were these deprecations added in the 1st place and why are they being removed now?

These deprecations were added as part of deprecating the OAuth 1.0[1] and with this PR some of those deprecations were removed as we identified some usages of those classes. 

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1447

PR #1495, opened 2020-11-10
clear cache when token revocation for authzCode reuse

janakamarasena:
Resolves https://github.com/wso2/product-is/issues/10401


IsuraD (approved):

pulasthi7 (approved):

mefarazath:
Shall we also improve our integration tests to do an API call or user info call after authz code reuse?

mefarazath (approved):

PR #1504, opened 2020-11-18
add required permissions and scopes

sherenewso2:

Resolves https://github.com/wso2/product-is/issues/10520

### Follow up actions

- [ ] Update docs oauth2-scope-management-rest-apis.md file `url` param to raw link of original yaml file.


components/org.wso2.carbon.identity.oauth.scope.endpoint/src/main/resources/api.identity.oauth2.scope.endpoint.yaml
```diff
@@ -35,8 +35,8 @@ host: localhost:9443
 # Will be prefixed to all paths.
 
 # For a tenant, basepath should be /t/{tenant_domain}/api/identity/oauth2/v1.0.
```
DMHP, line 37: If we are to provide the tenant path as the base path by default we can get rid of this comment.

ashendes (approved):

PR #1510, opened 2020-12-05
Reuse existing refresh token when issuing a new access token

rksk:
Resolves https://github.com/wso2/product-is/issues/10684

This issue has been introduced while fixing https://github.com/wso2/product-is/issues/4876 since we had not passed `exisitngTokenBean` to `generateNewAccessToken()` to avoid it getting marked as EXPIRED while it is already REVOKED.

Hence, a new parameter called `markExistingTokenExpired` was added for `generateNewAccessToken()` and passed the `exisitngTokenBean` in both of the following scenarios.
- Revoke existing and issue a new token; `markExistingTokenExpired` is false since we have already revoked the token
- Issue a new token since the existing one is expired; `markExistingTokenExpired` is true

Inside `generateNewAccessToken()` we are generating the `newTokenBean` and it will include the refresh token from `exisitngTokenBean` if it is not already expired.

But  the `exisitngTokenBean` is passed to  `persistAccessTokenInDB()` only if `markExistingTokenExpired` is true since it does the token expiry task which is not required when `markExistingTokenExpired` is true.

-

can merge this pull request.]

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
@@ -162,7 +162,7 @@ public abstract class AbstractAuthorizationGrantHandler implements Authorization
                 log.debug("No active access token found for client Id: " + consumerKey + ", user: " +
                         authorizedUser + " and scope: " + scope + ". Therefore issuing new token.");
             }
-            return generateNewAccessToken(tokReqMsgCtx, scope, consumerKey, existingTokenBean,
+            return generateNewAccessToken(tokReqMsgCtx, scope, consumerKey, existingTokenBean, true,
                     oauthTokenIssuer);
         }
     }
@@ -337,7 +337,8 @@ public abstract class AbstractAuthorizationGrantHandler implements Authorization
                         .TOKEN_STATE_REVOKED);
         clearExistingTokenFromCache(tokReqMsgCtx, existingTokenBean);
 
-        return generateNewAccessToken(tokReqMsgCtx, scope, consumerKey, null, oauthTokenIssuer);
+        return generateNewAccessToken(tokReqMsgCtx, scope, consumerKey, existingTokenBean, false,
+                oauthTokenIssuer);
     }
 
     private OAuth2AccessTokenRespDTO issueExistingAccessToken(OAuthTokenReqMessageContext tokReqMsgCtx, String scope,
@@ -351,6 +352,7 @@ public abstract class AbstractAuthorizationGrantHandler implements Authorization
 
     private OAuth2AccessTokenRespDTO generateNewAccessToken(OAuthTokenReqMessageContext tokReqMsgCtx, String scope,
                                                             String consumerKey, AccessTokenDO existingTokenBean,
+                                                            boolean expireExistingToken,
                                                             OauthTokenIssuer oauthTokenIssuer)
             throws IdentityOAuth2Exception {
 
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -351,6 +352,7 @@ private OAuth2AccessTokenRespDTO issueExistingAccessToken(OAuthTokenReqMessageCo
 
     private OAuth2AccessTokenRespDTO generateNewAccessToken(OAuthTokenReqMessageContext tokReqMsgCtx, String scope,
                                                             String consumerKey, AccessTokenDO existingTokenBean,
+                                                            boolean markExistingTokenExpired,
```
ShanChathusanda93, line 355: Shall we change this variable to expireExistingToken? WDYT?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -360,9 +362,15 @@ private OAuth2AccessTokenRespDTO generateNewAccessToken(OAuthTokenReqMessageCont
         AccessTokenDO newTokenBean = createNewTokenBean(tokReqMsgCtx, oAuthAppBean, existingTokenBean, timestamp,
                 validityPeriodInMillis, oauthTokenIssuer);
         setDetailsToMessageContext(tokReqMsgCtx, validityPeriodInMillis, newTokenBean, timestamp);
-        // Persist the access token in database
-        persistAccessTokenInDB(tokReqMsgCtx, existingTokenBean, newTokenBean, timestamp,
-                newTokenBean.getAccessToken());
+        if (markExistingTokenExpired) {
```
ShanChathusanda93, line 365: Shall we add a comment to describe the if condition? Similar one to the following.
`From the below if condition, it will check whether the existing token needs to be expired or not. Based on that the corresponding parameters will be sent to the persistAccessTokenInDB method.`
WDYT?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -360,9 +362,15 @@ private OAuth2AccessTokenRespDTO generateNewAccessToken(OAuthTokenReqMessageCont
...
+            // Persist the access token in database and mark the existing token as expired
+            persistAccessTokenInDB(tokReqMsgCtx, existingTokenBean, newTokenBean, timestamp,
+                    newTokenBean.getAccessToken());
+        } else {
+            // Persist the access token in database
```
ShanChathusanda93, line 370: Shall we add a comment to describe the if condition? Similar one to the following.
`Persist the access token in database. Here the existingTokenBean is already updated in the database as revoked. Because of that we are sending the existingTokenBean as a null.`
WDYT?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -360,9 +362,19 @@ private OAuth2AccessTokenRespDTO generateNewAccessToken(OAuthTokenReqMessageCont
...
-                newTokenBean.getAccessToken());
+
+        /* Check whether the existing token needs to be expired and send the corresponding parameters
+        to the persistAccessTokenInDB method. */
+        if (expireExistingToken) {
+            // Persist the access token in database and mark the existing token as expired
+            persistAccessTokenInDB(tokReqMsgCtx, existingTokenBean, newTokenBean, timestamp,
+                    newTokenBean.getAccessToken());
+        } else {
+            // Persist the access token in database without updating the existingTokenBean
```
janakamarasena, line 373: lets the change comment to 
```
// Persist the access token in database without updating the existing token.
// The existing token should already be updated by this point.
```
We need to say that `expireExistingToken == false` should only be when the existing token is already updated in the db to a non active state.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -360,9 +362,19 @@ private OAuth2AccessTokenRespDTO generateNewAccessToken(OAuthTokenReqMessageCont
...
                 validityPeriodInMillis, oauthTokenIssuer);
         setDetailsToMessageContext(tokReqMsgCtx, validityPeriodInMillis, newTokenBean, timestamp);
-        // Persist the access token in database
-        persistAccessTokenInDB(tokReqMsgCtx, existingTokenBean, newTokenBean, timestamp,
-                newTokenBean.getAccessToken());
+
+        /* Check whether the existing token needs to be expired and send the corresponding parameters
+        to the persistAccessTokenInDB method. */
+        if (expireExistingToken) {
+            // Persist the access token in database and mark the existing token as expired
```
janakamarasena, line 369: Add a fullstop at the end.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
janakamarasena: Also lets add a debug log to check which condition gets through.

ShanChathusanda93 (approved):

janakamarasena (approved):

PR #1518, opened 2020-12-16
Replace placeholder values of the callback URL with server configurations

ayshsandu:

- A Util method the process a given templated URL(refer following example) to construct with server configurations 
     ```
       Templated URL = ${carbon.protocol}://${carbon.host}:${carbon.management.port}/some-path
       Processed URL = https://localhost:9443/some-path
       Processed URL when a ProxyPort is configured = https://localhost/some-path
     ```
- Whenever a callback URL validation is happening process the callback URL value to get the actual URL.
- Fix Related to https://github.com/wso2/product-is/issues/10514

Immediately.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
@@ -230,7 +230,7 @@ public class OAuth2Service extends AbstractAdmin {
      */
     private boolean validateCallbackURI(String callbackURI, OAuthAppDO oauthApp) {
         String regexp = null;
-        String registeredCallbackUrl = oauthApp.getCallbackUrl();
+        String registeredCallbackUrl = OAuth2Util.processURI(oauthApp.getCallbackUrl());
         if (registeredCallbackUrl.startsWith(OAuthConstants.CALLBACK_URL_REGEXP_PREFIX)) {
             regexp = registeredCallbackUrl.substring(OAuthConstants.CALLBACK_URL_REGEXP_PREFIX.length());
         }
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
@@ -45,6 +45,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.axiom.om.OMElement;
 import org.apache.axiom.om.impl.builder.StAXOMBuilder;
 import org.apache.axiom.util.base64.Base64Utils;
+import org.apache.axis2.engine.AxisConfiguration;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.collections.CollectionUtils;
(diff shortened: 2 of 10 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -230,7 +230,7 @@ public OAuth2ClientValidationResponseDTO validateClientInfo(String clientId, Str
      */
     private boolean validateCallbackURI(String callbackURI, OAuthAppDO oauthApp) {
         String regexp = null;
-        String registeredCallbackUrl = oauthApp.getCallbackUrl();
+        String registeredCallbackUrl = OAuth2Util.processURI(oauthApp.getCallbackUrl());
```
mefarazath, line 233: The callback URI is usually a specific endpoint to an app. Do we need to resolve placeholders for this?

mefarazath:
@ayshsandu callback URIs are usually specific endpoints to an application. Ideally, they shouldn't have any placeholders?
Are we making these changes considering system apps?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
ayshsandu: > @ayshsandu callback URIs are usually specific endpoints to an application. Ideally, they shouldn't have any placeholders?
> Are we making these changes considering system apps?

Yes, this change is done for the system apps. Change is introduced in https://github.com/wso2/identity-apps/pull/1452 
For other apps, templated values are not accepted.

ayshsandu:
> @ayshsandu callback URIs are usually specific endpoints to an application. Ideally, they shouldn't have any placeholders?
> Are we making these changes considering system apps?

Yes, this change is done for the system apps. Change is introduced in https://github.com/wso2/identity-apps/pull/1452 
For other apps, templated values are not accepted.

janakamarasena:
Agreed with @mefarazath. Since this is also one major validation for OAuth applications we shouldn't make this dynamic. Since the problematic cases are the portal apps can't we target them and do something like validating the URL at the startup and replacing the value. And not altering the URL at validations.  The URL should be static so having this dynamic capability could result in issues in the future.

ayshsandu:
> Agreed with @mefarazath. Since this is also one major validation for OAuth applications we shouldn't make this dynamic. Since the problematic cases are the portal apps can't we target them and do something like validating the URL at the server startup and replacing the value. And not altering the URL at validations. The URL should be static so having this dynamic capability could result in issues in the future.

@janakamarasena as explained in [this comment](https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1518#issuecomment-746156865), the templating is only available for system apps (myaccount and console) at the time they are being created for the first time. For any other application, templated value for a callback is not accepted. Already, there is a validation.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -3936,4 +3937,37 @@ public static String getInternalEveryoneRole(AuthenticatedUser user) throws Iden
...
+            if (mgtTransportPort <= 0) {
+                mgtTransportPort = CarbonUtils.getTransportPort(axisConfiguration, mgtTransport);
+            }
+            // If it's well known HTTPS port, skip adding port
+            if (mgtTransportPort == IdentityCoreConstants.DEFAULT_HTTPS_PORT) {
+                url = StringUtils.replace(url, ":" +
+                        IdentityConstants.CarbonPlaceholders.CARBON_PORT, "");
+            }
+
+        }
```
mefarazath, line 3965: We can move this to a private method and use it within the if-block that checks for place holders

janakamarasena:
@ayshsandu at which level are we doing the validation for the url? Also another downside i see is if we add a new feature that requires callback url validation then we will need to remember and call this method.

ayshsandu:
Decided not to fix this in the code level. Instead, advice users to manually change the URL of SYSTEM APPS when a hostname change or port change is happening.

PR #1526, opened 2021-01-13
Remove the access token from cache when revoking the token

nilasini:

Resolves https://github.com/wso2/product-is/issues/10932

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
@@ -25,6 +25,7 @@ import org.apache.commons.logging.LogFactory;
 import org.wso2.carbon.identity.base.IdentityConstants;
 import org.wso2.carbon.identity.base.IdentityException;
 import org.wso2.carbon.identity.core.util.IdentityUtil;
+import org.wso2.carbon.identity.oauth.OAuthUtil;
 import org.wso2.carbon.identity.oauth.cache.OAuthCache;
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
```diff
@@ -258,6 +260,16 @@ private AuthzCodeDO getPersistedAuthzCode(OAuth2AccessTokenReqDTO tokenReqDTO) t
                 //revoking access token issued for authorization code as per RFC 6749 Section 4.1.2
                 revokeExistingAccessTokens(validationResult.getTokenId(), validationResult.getAuthzCodeDO());
                 clearTokenCache(tokenAlias, validationResult.getTokenId());
+
```
mefarazath, line 263: Can we use the already existing utils as used in https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/e0c9eeba4d5059864d86a32e34e83fb86ebecd84/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java#L419-L431 ?

On a separate note, can you see if we can reduce the number of utils and have a single method to handle token revocation cache clearing for a token so that it will be useful in other places.

nilasini: Used the existing utils. Will send a separate PR to reduce the number of utils and have a single method.

nilasini: Tracking through https://github.com/wso2/product-is/issues/10959

mefarazath (approved):

PR #1546, opened 2021-02-10
Return null if token found from cache is expired

mefarazath:


rmsamitha:
Fixes https://github.com/wso2/product-apim/issues/9909

IsuraD (approved):

PR #1549, opened 2021-02-16
Skip generating JWT during the token introspection if 'required_claim' parameter is not sent

DinikaSen:
The introspection request can have an additional parameter `required_claims` to obtain a JWT with required claims along with the  token validation response[1]. However currently this JWT generation happens without checking whether the `required_claims` parameter is set in the introspection request or not. Generation this JWT is an additional overhead to the token introspection flow when the introspection is called without `required_claims` optional parameter.

Also this PR introduces a new attribute to store the authorized user object as the existing username attribute may not have the domain appended to it, if the subject identifier is built based in the SP config[2].

Resolves : https://github.com/wso2/product-is/issues/11257

[1] https://github.com/wso2/docs-is/issues/1373
[2] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/6f5f6caa9f0c4a3724141d0bfb5a786bdaa7f727/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java#L525

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/introspection/OAuth2IntrospectionEndpoint.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/introspection/OAuth2IntrospectionEndpoint.java
@@ -84,10 +84,10 @@ public class OAuth2IntrospectionEndpoint {
                     entity("{\"error\": \"" + INVALID_INPUT + "\"}").build();
         }
 
-        String[] claimsUris;
+        String[] claimsUris = null;
         if (StringUtils.isNotEmpty(requiredClaims)) {
             claimsUris = requiredClaims.split(",");
-        } else {
+        } else if (requiredClaims != null && requiredClaims.length() == 0) {
             claimsUris = new String[0];
         }
 
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dto/OAuth2IntrospectionResponseDTO.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dto/OAuth2IntrospectionResponseDTO.java
@@ -18,6 +18,8 @@
 
 package org.wso2.carbon.identity.oauth2.dto;
 
+import org.wso2.carbon.identity.application.authentication.framework.model.AuthenticatedUser;
+
 import java.util.HashMap;
 import java.util.Map;
 
@@ -113,6 +115,11 @@ public class OAuth2IntrospectionResponseDTO {
      * OPTIONAL. Authorized user type of the token. (APPLICATION or APPLICATION_USER)
      */
     private String aut;
+
+    /*
+     * OPTIONAL. User object for the resource owner who authorized this token.
+     */
+    private AuthenticatedUser authorizedUser;
 
     /*
      * this is used for extensions.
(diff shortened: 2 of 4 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/introspection/OAuth2IntrospectionEndpoint.java
```diff
@@ -84,10 +84,10 @@ public Response introspect(@FormParam("token") String token, @FormParam("token_t
                     entity("{\"error\": \"" + INVALID_INPUT + "\"}").build();
         }
 
-        String[] claimsUris;
+        String[] claimsUris = null;
         if (StringUtils.isNotEmpty(requiredClaims)) {
             claimsUris = requiredClaims.split(",");
-        } else {
+        } else if (requiredClaims != null && requiredClaims.length() == 0) {
```
GANGANI, line 90: What is the reason of having requiredClaims.length() == 0?

DinikaSen: Sending empty value in required_claims should return a JWT with no custom claims within the introspection response according to the description here (wso2/docs-is#1373). In order to find whether the required_claims param is sent with the request, we use this `claimsUris` string array (If it is null, required_claims param is not sent, if it is not null, required_claims param is sent either with a value or without a value) . As we need send a JWT when required_claims is an empty value, we need to set `claimsUris` array to an empty array.

GANGANI: Shall we use StringUtils.isNotBlank(requiredClaims)?

DinikaSen: Here we need to check both the conditions `requiredClaims != null` and `requiredClaims.length() == 0`. If we get `requiredClaims=""`, this check should return `true`. But StringUtils.isNotBlank("")  = false. Therefore, we cannot use StringUtils.isNotBlank(requiredClaims) here.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dto/OAuth2IntrospectionResponseDTO.java
```diff
@@ -114,6 +116,11 @@
      */
     private String aut;
 
+    /*
+     * OPTIONAL. User object for the resource owner who authorized this token.
```
mefarazath, line 120: Represents the resource owner who authorized this token.

mefarazath: BTW what would be the behaviour for client credential grant type?

mefarazath (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dto/OAuth2IntrospectionResponseDTO.java
DinikaSen: Done with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1549/commits/0c7f5f1efe5bc0667cf986d6c4bc360f6c91f99c

PR #1553, opened 2021-02-24
Handle UserStoreClientException in password grant

rksk:
Related to wso2/product-is#11332

This PR is to convey the message to the end-user if the UserStoreClientException is thrown from the userstore level.


ashensw:
@rksk Build is failing with a [COMPILATION ERROR](https://github.com/wso2/product-is/runs/2009755759?check_suite_focus=true). Seems we have to bump the kernel version. Please have a look.

ashensw (approved):

PR #1555, opened 2021-03-02
Improve requested claims consent experience

GANGANI:
### Description

Post consent handling should do for both the approved and denied claims. Hense removed the condition `isPostConsentHandlingRequired(approvedClaimIds)`

Resolves: https://github.com/wso2/product-is/issues/9009, https://github.com/wso2-enterprise/asgardeo-product/issues/2144
Related PRs: https://github.com/wso2/carbon-consent-management/pull/180,https://github.com/wso2/carbon-identity-framework/pull/3413

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
@@ -511,35 +511,29 @@ public class OAuth2AuthzEndpoint {
             }
 
             List<Integer> approvedClaimIds = getUserConsentClaimIds(oAuthMessage);
-            if (isPostConsentHandlingRequired(approvedClaimIds)) {
-                serviceProvider = getServiceProvider(clientId);
-                /*
-                    With the current implementation of the SSOConsentService we need to send back the original
-                    ConsentClaimsData object we got during pre consent stage. Currently we are repeating the API call
-                    during post consent handling to get the original ConsentClaimsData object (Assuming there is no
-                    change in SP during pre-consent and post-consent).
 
-                    The API on the SSO Consent Service will be improved to avoid having to send the original
-                    ConsentClaimsData object.
-                 */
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
```diff
@@ -511,35 +511,29 @@ private void handlePostConsent(OAuthMessage oAuthMessage) throws ConsentHandling
             }
 
             List<Integer> approvedClaimIds = getUserConsentClaimIds(oAuthMessage);
-            if (isPostConsentHandlingRequired(approvedClaimIds)) {
```
tharindu-bandara, line 514: Just wondering what will happen if we have already approved some of the consent. Ex: Claims A, B are requested and already approved. Then later Claim C is requested. Will it ask only for Claim C in the next authentication? By the looks of the code shouldn't we compare already approved claims and requesting claims and then decide `isPostConsentHandlingRequired ` ?

GANGANI: As you said, when claim C is requested, it will only ask the consent for claim C and it will handle trough the pre consent and here it handles the post consent. With the `getUserConsentClaimIds(oAuthMessage)` method it will get the approved claim Ids to pass to the `processConsent` method[1] to process consent and store the approved and denied consents accordingly. The `isPostConsentHandlingRequired(approvedClaimIds)` condition will true only if there is an approvedClaimIds. Hence removed that condition in order to the claims that are denied also.

Let's say  Claims A, B are requested and already approved. Then later Claim C is requested and didn't give consent. Then according to the fix done at [2], in the "CM_SP_PURPOSE_PII_CAT_ASSOC" table, we store it as a consent denied claim. Hence `(isPostConsentHandlingRequired(approvedClaimIds))` with this condition that denied consent storing part will be ignored and that's the reason to remove that condition.

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1555/files#diff-debebfb0b01b553afd77f73d24e1df466030d3d19b218e0d7b3005364a05939aR535
[2] https://github.com/wso2/carbon-consent-management/pull/180

tharindu-bandara (approved):

PR #1558, opened 2021-03-08
Improve OIDCSessionDataCache to be a persistence cache

sachiniWettasinghe:
Resolves https://github.com/wso2/product-is/issues/11313

OIDCSessionDataCache is not implemented as a persistence cache. This causes OIDC logout fails in a multi-node environment if sticky sessions are not enabled. Hence, with this PR the OIDCSessionDataCache is improved to be a persistence cache to overcome this issue.

### DIFF
--- a/components/org.wso2.carbon.identity.oidc.session/src/test/java/org/wso2/carbon/identity/oidc/session/cache/OIDCSessionDataCacheTest.java
+++ b/components/org.wso2.carbon.identity.oidc.session/src/test/java/org/wso2/carbon/identity/oidc/session/cache/OIDCSessionDataCacheTest.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


components/org.wso2.carbon.identity.oidc.session/src/test/java/org/wso2/carbon/identity/oidc/session/cache/OIDCSessionDataCacheTest.java
```diff
@@ -0,0 +1,60 @@
...
+public class OIDCSessionDataCacheTest {
+
+    @Test
+    public void testGetInstance() {
+
+        assertNotNull(OIDCSessionDataCache.getInstance(), "OIDCSessionDataCache is null.");
+    }
+
+    @Test
+    public void testAddToCache() {
```
mefarazath, line 41: Can we add a couple of test cases to cover the scenario where a cache miss happens at the in memory cache and the data is fetched from the session data store?

sachiniWettasinghe: Fixed by 05697c21b56abf3d790d8577ef4e072b03f2922a

ashensw (approved):

sachiniWettasinghe:
When built locally, the build passes successfully and the PR builder has also passed. Hence merging the PR.

PR #1570, opened 2021-03-23
Add dcr and oauth scope APIs as jars inside the identity rest dispatcher

senthalan:
Add dcr and oauth scope as jars and deploy them as jars inside API rest dispatcher 
Related to https://github.com/wso2/product-is/issues/11425

As `api#identity#oauth2#dcr#v1.1.war` and `api#identity#oauth2#v1.0.war` are packed by other products, those wars will be kept as it is in the features and new jars will be copied inside `api_resources` then moved to api-dispatcher and wars will be removed in the product-is. Other products can remove `api_resources` folder and keep the wars.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.dcr.endpoint/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth.dcr.endpoint/pom.xml
@@ -175,6 +175,36 @@
                 </configuration>
             </plugin-->
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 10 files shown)


mefarazath (commented):
Since this is the master branch, can't we announce this publicly and do a version bump to reflect the major change?

senthalan:
> Since this is the master branch, can't we announce this publicly and do a version bump to reflect the major change?

Sent out mail to iam-dev@wso2.org

PR #1571, opened 2021-03-25
Set authorized internal scopes to a property in context

vihanga-liyanage:
- Previously internal scopes (scopes with the prefix "internal_") were removed before invoking the scope validators. With this change, authorized internal scopes are set to a property in the context, so that they can be seen by the scope validators if needed.

Refers https://github.com/wso2-enterprise/asgardeo-product/issues/2631

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/Oauth2ScopeConstants.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/Oauth2ScopeConstants.java
@@ -30,7 +30,6 @@ public class Oauth2ScopeConstants {
     public static final String SCOPE_TYPE_OAUTH2 = "OAUTH2";
     public static final String SCOPE_TYPE_OIDC = "OIDC";
 
-
     /**
      * Enums for error messages.
      */
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
@@ -298,6 +298,9 @@ public class AccessTokenIssuer {
         // the other scope validators.
         removeInternalScopes(tokReqMsgCtx);
 
+        // Adding the authorized internal scopes to tokReqMsgCtx for any special validators to use.
+        tokReqMsgCtx.setAuthorizedInternalScopes(authorizedInternalScopes);
+
         boolean isDropUnregisteredScopes = OAuthServerConfiguration.getInstance().isDropUnregisteredScopes();
         if (isDropUnregisteredScopes) {
             if (log.isDebugEnabled()) {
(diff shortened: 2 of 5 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -298,6 +299,9 @@ public OAuth2AccessTokenRespDTO issue(OAuth2AccessTokenReqDTO tokenReqDTO)
         // the other scope validators.
         removeInternalScopes(tokReqMsgCtx);
 
+        // Adding the authorized internal scopes to tokReqMsgCtx as a property for any special validators to use.
+        tokReqMsgCtx.addProperty(AUTHORISED_INTERNAL_SCOPES_PARAM_NAME, authorizedInternalScopes);
```
mefarazath, line 303: Instead of a property how about making this a first class attribute of the context?

janakamarasena (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/Oauth2ScopeConstants.java
```diff
@@ -30,6 +30,7 @@
     public static final String SCOPE_TYPE_OAUTH2 = "OAUTH2";
     public static final String SCOPE_TYPE_OIDC = "OIDC";
 
+    public static final String AUTHORISED_INTERNAL_SCOPES_PARAM_NAME = "AUTHORISED_INTERNAL_SCOPES";
```
mefarazath, line 33: AUTHORISED or AUTHORIZED ??

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
vihanga-liyanage: I guess will have to change the model and everything. Haven't we introduced this property field for this sort of modifications?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/Oauth2ScopeConstants.java
vihanga-liyanage: Both are correct I guess. Z is used in US English. :D

mefarazath: Which one do we usually use?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
mefarazath: TokenContext is a model used internally I guess (Not exposed directly via an admin service). For readability, I prefer to have it as a separate attribute in the context.

Otherwise we have to document this somewhere. Otherwise extension developers won't realize the internal scopes are available unless they debug and check the context properties.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/Oauth2ScopeConstants.java
dewniMW: In docs we use "Z"

mefarazath (approved):

PR #1587, opened 2021-04-29
Fix the issue of not having the newly added associate tenant in ID token

GANGANI:
### Description
The associate tenants were added as requested claims in the console and that associated tenant list is expected in the ID token. But the issue was, when a new tenant is created, the new ID token obtained after the immediate refresh does not have the updated list of tenants.

The issue was due to not correctly updating the user attribute cache against the access token. With this fix, it will copy the attributes cached against code to token in the authorization code flow before building the ID token. 

Resolves: https://github.com/wso2-enterprise/asgardeo-product/issues/3368

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultOIDCClaimsCallbackHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultOIDCClaimsCallbackHandler.java
@@ -241,22 +241,18 @@ public class DefaultOIDCClaimsCallbackHandler implements CustomClaimsCallbackHan
     }
 
     private Map<ClaimMapping, String> getCachedUserAttributes(OAuthTokenReqMessageContext requestMsgCtx) {
-
-        Map<ClaimMapping, String> userAttributes =
-                getUserAttributesCachedAgainstAuthorizationCode(getAuthorizationCode(requestMsgCtx));
+        Map<ClaimMapping, String> userAttributes = getUserAttributesCachedAgainstToken(getAccessToken(requestMsgCtx));
         if (log.isDebugEnabled()) {
-            log.debug("Retrieving claims cached against authorization_code for user: " +
-                    requestMsgCtx.getAuthorizedUser());
+            log.debug("Retrieving claims cached against access_token for user: " + requestMsgCtx.getAuthorizedUser());
         }
         if (isEmpty(userAttributes)) {
             if (log.isDebugEnabled()) {
-                log.debug("No claims cached against the authorization_code for user: " +
-                        requestMsgCtx.getAuthorizedUser() +
-                        ". Retrieving claims cached against the access_token code.");
+                log.debug("No claims cached against the access_token for user: " + requestMsgCtx.getAuthorizedUser() +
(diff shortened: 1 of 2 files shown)


mefarazath:
Can we explain the issue and how this solves the issue?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultOIDCClaimsCallbackHandler.java
```diff
@@ -241,18 +241,22 @@ private boolean isPreserverClaimUrisInAssertion(OAuthTokenReqMessageContext requ
     }
 
     private Map<ClaimMapping, String> getCachedUserAttributes(OAuthTokenReqMessageContext requestMsgCtx) {
-        Map<ClaimMapping, String> userAttributes = getUserAttributesCachedAgainstToken(getAccessToken(requestMsgCtx));
```
mefarazath, line 244: Usually in the authorization code, once a token is issued the attributes cached against authorization code are stored against the token.

So ideal case, the same attributes stored against the authorization code will be available against the token at the point the token is id_token is built

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultOIDCClaimsCallbackHandler.java
```diff
@@ -241,18 +241,22 @@ private boolean isPreserverClaimUrisInAssertion(OAuthTokenReqMessageContext requ
...
         if (isEmpty(userAttributes)) {
             if (log.isDebugEnabled()) {
-                log.debug("No claims cached against the access_token for user: " + requestMsgCtx.getAuthorizedUser() +
-                        ". Retrieving claims cached against the authorization code.");
+                log.debug("No claims cached against the authorization_code for user: " +
+                        requestMsgCtx.getAuthorizedUser() +
+                        ". Retrieving claims cached against the access_token code.");
             }
-            userAttributes = getUserAttributesCachedAgainstAuthorizationCode(getAuthorizationCode(requestMsgCtx));
+            userAttributes = getUserAttributesCachedAgainstToken(getAccessToken(requestMsgCtx));
```
IsuraD, line 257: We have to give priority to access token..
if it presents get the attribute from access token...

else check authorization code presents and if exits get user attributes from authorization code.

mefarazath: +1

At the point of issuing an id_token in ideal case the attributes stored against code and token should be same. Since there is a step to copy attributes cached against code to token in the authorization code flow.

If the attributes cached against the token are not updated then the issue might be else where

GANGANI: yes. The issue was due to copying the attributes cached against code to token in the authorization code flow before building the ID token. Fixed it by moving the `addUserAttributesAgainstAccessToken` method before the ID token build.

ashensw (approved):

PR #1599, opened 2021-05-18
Use bang (!) operator

vordimous:

- This will evaluate to true if either the length is 0 or if length is null/undefined (meaning data is not a string)

as soon as possible, this causes a crash loop if e.data is an object

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oidc.session/src/main/resources/op_iframe.html
+++ b/components/org.wso2.carbon.identity.oidc.session/src/main/resources/op_iframe.html
@@ -10,7 +10,7 @@
             }
 
             var stat = '';
-            if (e.data === null || e.data == 'null' || e.data.length == 0) {
+            if (e.data === null || e.data == 'null' || !e.data.length) {
                 stat = 'error';
             } else {
                 var client_id = e.data.split(' ')[0];


components/org.wso2.carbon.identity.oidc.session/src/main/resources/op_iframe.html
```diff
@@ -10,7 +10,7 @@
             }
 
             var stat = '';
-            if (e.data === null || e.data == 'null' || e.data.length == 0) {
+            if (e.data === null || e.data == 'null' || !e.data.length) {
```
brionmario, line 13: Hi @vordimous,

Can't we just check if the type of `e.data` is a string?

```suggestion
            if (e.data === null || e.data == 'null' || typeof e.data === "string") {
```

vordimous: The original code checked for length, meaning the intention was to prevent empy string as well. Just checking for string wouldn't care if ut was an empty string.

brionmario: When we check if the type is a string before doing the split, we can avoid the `TypeError: e.data.split is not a function` AFAIS.

Can't we check `e.data === ''" || e.data === undefined` as well to address the empty string scenario? WDYT?

vordimous: `if (e.data === null || e.data == 'null' || typeof e.data !== "string" || e.data === ''" || e.data === undefined) {` make the code lest readable and maintainable. Considering the ! operator covers those cases plus others and is standard part of the ES spec. Unless we add all[ 4 checks ](https://levelup.gitconnected.com/understanding-weirdness-of-operator-in-javascript-b7609fcab810#:~:text=The%20!%20operation%20on%200%2C%20null%2C%20undefined%2C%20NaN%20returns%20true%20since%20they%20are%20falsey.)we are limiting the conditional statement.

brionmario: True. But the concert i had with the bang operator is that it'll pass for all the truthy values.
Ex:  `e.data.length === "1"` etc.

I would prefer if we can check if `length` is in the prototype and if length is <= 0.

```js
 if (e.data === null || e.data == 'null' || typeof e.data !== "string" || e.data.length <= 0) {
 ```
 
 Let me know if you think if this would work. Else, we can merge the PR.

vordimous: Sure thing. If there is a concern for truthy it may be best to change the order all together 

`if (typeof e.data !== "string" || e.data == 'null' || e.data.trim().length <= 0) {`

In this case we only will evaluate string conditions ('null' and length) in the case that we have a string. any other typeof (null and undefined included) will short circuit the condition. It also orders the conditions to be more clear what is expected in the `data` prop.

vordimous: Any thoughts @brionmario

GDRDABARERA (approved):

PR #1604, opened 2021-05-22
Support for Git based PR builder

omindu:
- Introducing git workflow to build the PR
- Renaming the old product-is PR builder
- Increasing spotbugs memory limits to prevent OOM issues


.github/workflows/product-is-pr-builder.yml
```diff
@@ -0,0 +1,73 @@
+name: Product IS PR Builder
```
janakamarasena, line 1: Lets remove this since we have it in product-is

.github/workflows/pr-builder.yml
```diff
@@ -1,73 +1,42 @@
...
+
+name: PR Builder
 
 on:
-  pull_request_target:
-    types: [labeled]
-    branches:
-      - main
-      - master
+  pull_request:
```
janakamarasena, line 9: I think this might be the more suitable one https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target since we want to build the forked repo

.github/workflows/pr-builder.yml
```diff
@@ -1,73 +1,42 @@
...
-        if-no-files-found: warn
-    - name: Archive product-is surefire reports
-      if: always()
-      uses: actions/upload-artifact@v2
-      with:
-        name: product-is-surefire-report
-        path: |
-          product-is/**/surefire-reports
-        if-no-files-found: warn
+        path: ~/.m2/repository
```
janakamarasena, line 34: Lets remove the caching for this

.github/workflows/pr-builder.yml
janakamarasena: Will have to add triggeres on `opened` and `edited` i guess

janakamarasena (approved):

PR #1658, opened 2021-08-20
Token introspection restricted to tenant specific admins.

ashendes:
Admins from different tenants are blocked by default from introspecting token of other tenants. This can be disabled by applying the following config.

```
[oauth]
allow_cross_tenant_token_introspection=true
```


pulasthi7 (approved):

ashendes:
This was reverted due to a similar fix already available in the tenant context valve level.

PR #1670, opened 2021-09-02
Feature dpop

kalanika:

-   Introduced new token type called DPoP
-    Updated JWT Token Issuer to add cnf claim to the DPoP tokens
-    Added new application config to handle dpop state (Enable, Disable or make it mandatory)
-     DPoP feature can be hide based on config
-     Updated introspection response to return cnf for DPoP tokens
-     DPoP token will be bind to a public key thumbprint value. If application is configured with other binding as SSO session or cookies then token binding type will be stored as comma separated. (temp solution as proper solution require a schema change)
-     Introduced error handler to handle DPoP related error
-     DPoP proof validity during the token generation is handled by config


kalanika:
Resolves wso2/product-is#9102

PR #1685, opened 2021-09-16
Change the default value for subject identifier

madurangasiriwardena:
Part of https://github.com/wso2/product-is/issues/12504

All the applications used after fix will default to using user id as the subject identifier.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
@@ -26,8 +26,16 @@ import org.apache.commons.logging.LogFactory;
 import org.apache.oltu.oauth2.common.error.OAuthError;
 import org.apache.oltu.oauth2.common.message.types.GrantType;
 import org.owasp.encoder.Encode;
+import org.wso2.carbon.identity.application.authentication.framework.exception.UserIdNotFoundException;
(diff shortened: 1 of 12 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -348,6 +360,13 @@ public OAuth2AccessTokenRespDTO issue(OAuth2AccessTokenReqDTO tokenReqDTO)
             // set the token request context to be used by downstream handlers. This is introduced as a fix for
             // IDENTITY-4111.
             OAuth2Util.setTokenRequestContext(tokReqMsgCtx);
+
+            AuthenticatedUser authorizedUser = tokReqMsgCtx.getAuthorizedUser();
+            if (authorizedUser.getAuthenticatedSubjectIdentifier() == null) {
+                authorizedUser.setAuthenticatedSubjectIdentifier(
```
mefarazath, line 366: Will there be a performance impact for grant types like client_credentials which usually do not load the SP to determine the sub value?

mefarazath: A suggestion from me is to load only the part of the SP related to subject attribute instead of the whole SP even in the worst case.

mefarazath (approved):

mefarazath:
@madurangasiriwardena does this fix cover the implicit and hybrid flows too?

madurangasiriwardena:
@mefarazath, yes, both the flows are verified.

PR #1697, opened 2021-09-28
Validation for PKCE Parameters inside the JWT Request object

Ashi1993:
# Purpose
According to the PKCE spec[1], an IAM Component supporting PKCE is supposed to accept several additional parameters in the authorization request (code_challenge, code_challenge_method) and token request (code_verifier). But the specification does not provide a clear-cut definition on where to send these parameters. Only in the Appendix B example indicates that these parameters could be sent as query params to the Authorization request. WSO2 Identity Server also supports the same parameters only as query params in Authorize request.

Since there is no clear indication on where to send the parameters, PKCE parameters for the authorization request (code_challenge, code_challenge_method) can be passed inside the JWT Request object. But current implementation does not support PKCE parameters inside the JWT Request object.
 
To fix the issue, 
- override relevant PKCE parameters from the request object to OAuth2Parameters (if PKCE parameters are present in request object) 
- PKCE parameters are set to the OAuth2Parameters (after PKCE parameter validation) before handling OIDC request object

[1] - https://datatracker.ietf.org/doc/html/rfc7636

**Issue** : https://github.com/wso2/product-is/issues/12371

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
@@ -1627,26 +1627,9 @@ public class OAuth2AuthzEndpoint {
         }
 
         if (isPkceSupportEnabled()) {
-            String pkceChallengeCode;
-            String pkceChallengeMethod;
+            String pkceChallengeCode = getPkceCodeChallenge(oAuthMessage, params);
+            String pkceChallengeMethod = getPkceCodeChallengeMethod(oAuthMessage, params);
 
-            //If the code_challenge is in the request object, then it is added to Oauth2 params before this point
-            if (params.getPkceCodeChallenge() != null) {
-                //If Oauth2 params contains code_challenge get value from Oauth2 params
-                pkceChallengeCode = params.getPkceCodeChallenge();
-            } else {
-                //Else retrieve from request query params
-                pkceChallengeCode = oAuthMessage.getOauthPKCECodeChallenge();
-            }
-
-            //If the code_challenge_method is in the request object, then it is added to Oauth2 params before this point
-            if (params.getPkceCodeChallengeMethod() != null) {
-                //If Oauth2 params contains code_challenge_method get value from Oauth2 params
-                pkceChallengeMethod = params.getPkceCodeChallengeMethod();
-            } else {
-                //Else retrieve from request query params
-                pkceChallengeMethod = oAuthMessage.getOauthPKCECodeChallengeMethod();
-            }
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
```diff
@@ -1625,8 +1627,26 @@ private String populateOauthParameters(OAuth2Parameters params, OAuthMessage oAu
...
+            String pkceChallengeMethod;
+
+            //If the code_challenge is in the request object, then it is added to Oauth2 params before this point
+            if (params.getPkceCodeChallenge() != null) {
+                //If Oauth2 params contains code_challenge get value from Oauth2 params
+                pkceChallengeCode = params.getPkceCodeChallenge();
+            } else {
+                //Else retrieve from request query params
+                pkceChallengeCode = oAuthMessage.getOauthPKCECodeChallenge();
+            }
```
janakamarasena, line 1640: Lets add this to a private method

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
```diff
@@ -1625,8 +1627,26 @@ private String populateOauthParameters(OAuth2Parameters params, OAuthMessage oAu
...
+
+            //If the code_challenge_method is in the request object, then it is added to Oauth2 params before this point
+            if (params.getPkceCodeChallengeMethod() != null) {
+                //If Oauth2 params contains code_challenge_method get value from Oauth2 params
+                pkceChallengeMethod = params.getPkceCodeChallengeMethod();
+            } else {
+                //Else retrieve from request query params
+                pkceChallengeMethod = oAuthMessage.getOauthPKCECodeChallengeMethod();
+            }
```
janakamarasena, line 1649: Lets add this to a private method

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
Ashi1993: moved to a private method

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
Ashi1993: moved to a private method

janakamarasena (approved):

PR #1707, opened 2021-10-04
fix identify authorized user subject identifier

somindatommy:
- Get the idp name instead of the Idp id so that the user id can be resolved from the subject identifier.
- Set 'username' as the `subjectIdentifier` for the federated users since the user id is resolved from the `IDN_AUTH_USER` table.


emswbandara (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -3556,6 +3558,7 @@ public static AuthenticatedUser createAuthenticatedUser(String username, String
             } else {
                 authenticatedUser.setFederatedIdPName(OAuth2Util.getFederatedIdPFromDomain(userStoreDomain));
             }
+            authenticatedUser.setUserId(getUserIdOfFederatedUser(username, tenantDomain, idpName));
```
emswbandara, line 3561: will the scenario work as expected for a local user? don't we need to set the user ID for a local user as well?

somindatommy: For local users, we have a different way of resolving the user id and it's handled separately.

piraveena (approved):

madurangasiriwardena (approved):

PR #1721, opened 2021-10-11
Add binding type validations for Application Management REST API

ChanikaRuchini:

> Add validations for the binding type.

Fixes - [#11219](https://github.com/wso2/product-is/issues/11219)


senthalan:
We support registering custom token binders as OSGi service, so this validation should not check against predefined names. This should check whether there is a token binder registered with this binding type. We can get the registers token binding implementations from `OAuth2ServiceComponentHolder.getInstance().getTokenBinder(bindingType);`

PR #1729, opened 2021-10-28
Remove client error logs

VivekVinushanth:
This PR removes logging client errors.


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpoint.java
```diff
@@ -153,7 +153,9 @@ private CarbonOAuthTokenRequest handleInvalidRequest(OAuthProblemException e)
                 log.debug("Error: " + e.getError() + ", description: " + e.getDescription());
             }
         } else {
```
mefarazath, line 155: We can remove the else block and move the logging outside

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/util/EndpointUtil.java
```diff
@@ -1199,9 +1199,14 @@ public static String getSPTenantDomainFromClientId(String clientId) {
         try {
             OAuthAppDO oAuthAppDO = OAuth2Util.getAppInformationByClientId(clientId);
             return OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO);
-        } catch (IdentityOAuth2Exception | InvalidOAuthClientException e) {
+        } catch (IdentityOAuth2Exception e) {
             log.error("Error while getting oauth app for client Id: " + clientId, e);
```
mefarazath, line 1203: If we are doing a fallback here, shouldn't this be a WARN or a DEBUG log?

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpoint.java
VivekVinushanth: The  logging has to be different as per the existing logic. Hence, else part is needed IMO.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2ScopeService.java
```diff
@@ -139,10 +139,15 @@ public Scope registerScope(Scope scope) throws IdentityOAuth2ScopeException {
...
-        } catch (IdentityOAuth2Exception | InvalidOAuthClientException e) {
+        } catch (IdentityOAuth2Exception e) {
             log.error("Error while getting oauth app for client Id: " + clientId, e);
             throw Oauth2ScopeUtils.generateServerException(Oauth2ScopeConstants.ErrorMessages.
                     ERROR_CODE_FAILED_TO_GET_ALL_SCOPES, e);
+        } catch (InvalidOAuthClientException e) {
+            if (log.isDebugEnabled()) {
+                log.debug("Error while getting oauth app for client Id: " + clientId, e);
+            }
+            return null;
```
senthalan, line 150: don't we have IdentityOAuth2ScopeClientException to through from here, so the API layer will handle it properly

VivekVinushanth: This is a public method and it throws only a server exception. So, though we can throw a client exception we might need to modify method signature which am bit skeptical about.

components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/util/EndpointUtil.java
VivekVinushanth: We are separating client and server errors. **IdentityOAuth2Exception**  is considered to be server error and we also catch **InvalidOAuthClientException** for which we will have debug logs.

senthalan (approved):

PR #1738, opened 2021-11-05
Modify the logic of adding at_hash claim to the id_token

ImalshaG:
The logic of adding the at_hash claim to the id_token is changed to support all the hybrid flow response types. 
The at_hash claim is added only if the access token is returned.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectSystemClaimImpl.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectSystemClaimImpl.java
@@ -18,6 +18,8 @@ package org.wso2.carbon.identity.openidconnect;
 import com.nimbusds.jose.JWSAlgorithm;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.io.Charsets;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.oltu.oauth2.common.message.types.ResponseType;
 import org.wso2.carbon.identity.oauth.common.OAuthConstants;
 import org.wso2.carbon.identity.oauth.config.OAuthServerConfiguration;
@@ -41,6 +43,8 @@ import static org.wso2.carbon.identity.oauth.common.OAuthConstants.OIDCClaims.C_
  * This class is used to inject system claims like c_hash, at_hash into the id_token.
  */
 public class OpenIDConnectSystemClaimImpl implements ClaimProvider {
+
+    public static final Log LOG = LogFactory.getLog(OpenIDConnectSystemClaimImpl.class);
     private static final String SHA384 = "SHA-384";
     private static final String SHA512 = "SHA-512";
     private JWSAlgorithm signatureAlgorithm = null;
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectSystemClaimImpl.java
```diff
@@ -145,8 +145,11 @@ private boolean isCodeHashApplicable(String responseType) {
 
     private boolean isAccessTokenHashApplicable(String responseType) {
         // At_hash is generated on an access token. Therefore check whether the response type returns an access_token.
-        // id_token and none response types don't return and access token
-        return !OAuthConstants.ID_TOKEN.equalsIgnoreCase(responseType) &&
-                !OAuthConstants.NONE.equalsIgnoreCase(responseType);
+        for (String respType : responseType.split(" ")) {
```
piraveena, line 148: can we create a new private method to get an array of all response types? If we have a separate method, it can be used for other flows as well.

WDYT?

ImalshaG: Added a new method.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectSystemClaimImpl.java
```diff
@@ -145,8 +145,15 @@ private boolean isCodeHashApplicable(String responseType) {
...
-                !OAuthConstants.NONE.equalsIgnoreCase(responseType);
+        for (String respType : getResponseTypes(responseType)) {
+            if (OAuthConstants.TOKEN.equalsIgnoreCase(respType)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private String[] getResponseTypes(String responseType) {
```
piraveena, line 156: formatting issue. should have a new line after method signature. Add the code style to your IDE

piraveena: can we do a null and blank check here? StringUtils.isNotBlank(responseType) and add a debug log if it is blank.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectSystemClaimImpl.java
```diff
@@ -145,8 +149,21 @@ private boolean isCodeHashApplicable(String responseType) {
...
+
+        if (!isNotBlank(responseType)) {
```
piraveena, line 162: shall we do splitting only if it is not blank? 
if(isNotBlank(responseType)){
return responseType.split(" ")
}else{
return an empty array
}

piraveena (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/OpenIDConnectSystemClaimImpl.java
```diff
@@ -145,8 +149,23 @@ private boolean isCodeHashApplicable(String responseType) {
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private String[] getResponseTypes(String responseType) {
+
+        if (isNotBlank(responseType)) {
+            return responseType.split(" ");
```
mefarazath, line 163: Just for the sake of argument can there be a case, where the response_type be separated by more than one space? or will it be considered as invalid at a layer above here?

mefarazath: something like
`response_type=code  id_token`

ImalshaG: It will be handled before coming to this point and will give the following response:
`https://localhost.com:9443/oauth2/a?error_description=invalid_request%2C+Invalid+response_type+parameter+value&state=Sr7tArZM9x&error=invalid_request`

mefarazath:
@ImalshaG lets link the related issue in the PR description

PR #1745, opened 2021-11-17
Set debug log instead of error log when trying with an invalid client id

GANGANI:
### Description

Set debug log instead of error log when trying to authenticate with an invalid client id as it is a client error.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
@@ -647,10 +647,9 @@ public class OAuth2Service extends AbstractAdmin {
             OAuthAppDO appDO = OAuth2Util.getAppInformationByClientId(consumerKey);
             return appDO.getState();
         } catch (IdentityOAuth2Exception | InvalidOAuthClientException e) {
-            String msg = "Error while finding application state for application with client_id: " + consumerKey;
-            log.error(msg);
             if (log.isDebugEnabled()) {
-                log.debug(msg, e);
+                log.debug("Error while finding application state for application with client_id: " +
+                        consumerKey, e);
             }
             return null;
         }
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -647,10 +647,9 @@ public String getOauthApplicationState(String consumerKey) {
             OAuthAppDO appDO = OAuth2Util.getAppInformationByClientId(consumerKey);
             return appDO.getState();
         } catch (IdentityOAuth2Exception | InvalidOAuthClientException e) {
-            String msg = "Error while finding application state for application with client_id: " + consumerKey;
-            log.error(msg);
             if (log.isDebugEnabled()) {
-                log.debug(msg, e);
+                log.debug("Error while finding application state for application with client_id: " +
```
ashendes, line 651: This could only be considered a client error in case its an `InvalidOAuthClientException`. It should be handled in the original manner in case its an `IdentityOAuth2Exception`. WDYT?

GANGANI: Fixed https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1745/commits/46410198e6640c719c0a7347ef3003a141e893cb

mefarazath (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -646,11 +646,13 @@ public String getOauthApplicationState(String consumerKey) {
         try {
             OAuthAppDO appDO = OAuth2Util.getAppInformationByClientId(consumerKey);
             return appDO.getState();
-        } catch (IdentityOAuth2Exception | InvalidOAuthClientException e) {
-            String msg = "Error while finding application state for application with client_id: " + consumerKey;
-            log.error(msg);
+        } catch (IdentityOAuth2Exception e) {
+            log.error("Error while finding application state for application with client_id: " + consumerKey, e);
+            return null;
+        } catch (InvalidOAuthClientException e) {
```
mefarazath, line 652: Does InvalidOAuthClientException extend IdentityOAuth2Exception?

GANGANI: No. Both these exceptions are extended from IdentityException

PR #1751, opened 2021-11-30
Fixed issue in validation redirect url with query params in Auth request 

Ashi1993:
# Purpose
When the redirect_uri contains query parameters DCR POST successfully registers the app, but the authorization flow fails with the "Registered callback does not match with the provided url." error. 

Redirect Url is saved with query parameter while creating the app without any change (Base64URL encoding) But in the authorization request it comes as URL encoded. Hence, when matching the URLs it returns an error. 

To fix this, we are replacing "?" in the stored URL as "\\?" to match with the URL in the authorization request

Related Issue : https://github.com/wso2/product-is/issues/12831


janakamarasena:
Lets add a unit test for this

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/OAuth2Service.java
```diff
@@ -233,7 +233,8 @@ private boolean validateCallbackURI(String callbackURI, OAuthAppDO oauthApp) {
         String regexp = null;
         String registeredCallbackUrl = oauthApp.getCallbackUrl();
         if (registeredCallbackUrl.startsWith(OAuthConstants.CALLBACK_URL_REGEXP_PREFIX)) {
-            regexp = registeredCallbackUrl.substring(OAuthConstants.CALLBACK_URL_REGEXP_PREFIX.length());
+            regexp = (registeredCallbackUrl.substring(OAuthConstants.CALLBACK_URL_REGEXP_PREFIX.length()))
+                    .replace("?", "\\?");
```
janakamarasena, line 237: Can we check if there are any inbuilt java methods to support this instead of a string replace.
Also what would happen if there is an actual regex containing `?` is here?
i.e. when registering the callback how do we differentiate whether the `?` is due to URL containing query params or due to being part of a regex pattern?

Ashi1993: Changed to replace only the first occurance of "?"

Ashi1993:
> Lets add a unit test for this

Added unit tests

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/OAuth2ServiceTest.java
```diff
@@ -705,6 +705,17 @@ public void testHandleAuthenticationFailureWithException() throws Exception {
...
     }
 
+    @Test
+    public void testValidateClientInfoWIthQueryParamsInCallbackURL() throws Exception {
+
+        String clientId = UUID.randomUUID().toString();
+        getOAuthAppDO(clientId, "dummyGrantType", "https://wso2.com?dummy1=1&dummy=2",
+                "carbon.super");
+        OAuth2ClientValidationResponseDTO oAuth2ClientValidationResponseDTO = oAuth2Service.
+                validateClientInfo(clientId, "https://wso2.com\\?dummy1=1&dummy=2");
```
mefarazath, line 715: Shouldn't we pass this without escapting here?

Ashi1993: The one comes in the auth request is similar to this

components/org.wso2.carbon.identity.oauth.dcr/src/main/java/org/wso2/carbon/identity/oauth/dcr/service/DCRMService.java
```diff
@@ -642,23 +642,29 @@ private boolean isRedirectURIMandatory(List<String> grantTypes) {
...
+        return regexPattern;
+    }
+
+    /**
+     * Method to escape query parameters in the redirect urls
+     * @param redirectURI
+     * @return
+     */
+    private String escapeQueryParamsIfPresent(String redirectURI) {
+        return redirectURI.replaceFirst("\\?", "\\\\?");
```
mefarazath, line 667:
```suggestion

        return redirectURI.replaceFirst("\\?", "\\\\?");
```

mefarazath (approved):

components/org.wso2.carbon.identity.oauth.dcr/src/test/java/org/wso2/carbon/identity/oauth/dcr/service/DCRMServiceTest.java
```diff
@@ -1054,4 +1058,89 @@ private void startTenantFlow() {
...
+                .substring(OAuthConstants.CALLBACK_URL_REGEXP_PREFIX.length());
+        for (String validCallback : validCallbackList) {
+            assertTrue(validCallback.matches(regexp));
+        }
+        for (String invalidCallback : invalidCallbackList) {
+            assertFalse(invalidCallback.matches(regexp));
+        }
+    }
+
+    private String createRegexPattern(List<String> redirectURIs) throws DCRMException {
```
pulasthi7, line 1129: This is the same method from the service it seems. Can you make the service method protected and re-use here?

Ashi1993: Done

pulasthi7 (approved):

PR #1812, opened 2022-03-15
Fixing case sensitive scope validation issue..

prasa7:
FIx for - "APIs cannot be invoked using a token which is generated with a scope based on userstore role"

Steps to reproduce.

Get APIM 2.6 pack and update to the latest or level 72.
Get WSO2 IS to be used as the userstore.
Start both IS and APIM server, go to the APIM Management console and configure an userstore.
Them create a role for that userstore (Example:TestRole)
Create a user with that user role.
Go to the publisher and create a new API
Create two scopes as TestRole and testRole. (To identify case sensitive issues)
Then assign those scopes to two different resources.
Generate a token and try to invoke API. One with TestRole would work and the other one with testRole would fail.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCScopeValidator.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCScopeValidator.java
@@ -396,7 +396,6 @@ public class JDBCScopeValidator extends OAuth2ScopeValidator {
             }
             return false;
         }
-
         boolean preservedCaseSensitive = Boolean.parseBoolean(System.getProperty("preservedCaseSensitive"));
 
         //Check if the user still has a valid role for this scope.
@@ -404,19 +403,16 @@ public class JDBCScopeValidator extends OAuth2ScopeValidator {
         if (!preservedCaseSensitive) {
             rolesOfScope.retainAll(Arrays.asList(userRoles));
         } else {
-
             for (String roleOfScope : rolesOfScope) {
                 rolesOfScope.remove(roleOfScope);
                 rolesOfScope.add(roleOfScope.toLowerCase());
             }
-
             ArrayList<String> userRolesLowercase = new ArrayList<>();
             for (String userRole : userRoles) {
                 userRolesLowercase.add(userRole.toLowerCase());
             }
             rolesOfScope.retainAll(userRolesLowercase);
         }
-
         rolesOfScope.retainAll(Arrays.asList(userRoles));
 
         if (rolesOfScope.isEmpty()) {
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCScopeValidator.java
```diff
@@ -397,8 +397,27 @@ private boolean isUserAuthorizedForScope(String scopeName, String[] userRoles, i
             return false;
         }
 
+
```
somindatommy: We can get rid of unnecessary new lines. WDYT?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCScopeValidator.java
```diff
@@ -397,8 +397,27 @@ private boolean isUserAuthorizedForScope(String scopeName, String[] userRoles, i
             return false;
         }
 
+	
+	boolean preservedCaseSensitive = Boolean.parseBoolean(System.getProperty("preservedCaseSensitive"));
+
```
somindatommy, line 401: We can get rid of unnecessary new lines. WDYT?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCScopeValidator.java
```diff
@@ -397,8 +397,27 @@ private boolean isUserAuthorizedForScope(String scopeName, String[] userRoles, i
...
         //Check if the user still has a valid role for this scope.
         Set<String> scopeRoles = new HashSet<>(rolesOfScope);
+        if (!preservedCaseSensitive) {
+            rolesOfScope.retainAll(Arrays.asList(userRoles));
+        } else {
+
```
somindatommy, line 407: We can get rid of unnecessary new lines. If there is not a considerable logic change, we do not need to add new lines. WDYT?

prasa7: Added the changes

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCScopeValidator.java
prasa7: Added the changes

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCScopeValidator.java
prasa7: Added the changes

somindatommy (approved):

ShanChathusanda93 (approved):

PR #1819, opened 2022-03-24
Check the requirement of having a configured public certificate or JWKS in SP to enable ID token encryption and request object signature validation

Maliesha:
Fix [wso2/product-is#11206](https://github.com/wso2/product-is/issues/11206)

Previously, if we enabled ID Token encryption without configuring the application certificate or the JWKS endpoint, it would give an internal server error when we attempt to generate an access token. 

As the solution, the certificate or JWKS configuration is mandatory if a user try to update or create an application. If the user try to enable ID token encryption or request object signature validation without configuring a public certificate or JWKS, we throw an exception with a specific error message.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
@@ -264,7 +264,8 @@ public class OAuthAdminServiceImpl {
 
                 ///Check the requirement of having a configured public certificate or JWKS in SP
                 boolean isIdTokenEncryptionEnabled = application.isIdTokenEncryptionEnabled();
-                boolean isRequestObjectSignatureValidationEnabled = application.isRequestObjectSignatureValidationEnabled();
+                boolean isRequestObjectSignatureValidationEnabled =
+                        application.isRequestObjectSignatureValidationEnabled();
                 handlePublicCertificateConfig(app.getOauthConsumerKey(), tenantDomain, isIdTokenEncryptionEnabled,
                         isRequestObjectSignatureValidationEnabled);
 
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
@@ -484,11 +484,11 @@ public class OAuthAdminServiceImpl {
         //Check the requirement of having a configured public certificate or JWKS in SP
         boolean isIdTokenEncryptionEnabled = !oauthappdo.isIdTokenEncryptionEnabled() &&
                 consumerAppDTO.isIdTokenEncryptionEnabled();
-        boolean isRequestObjectSignatureValidation = !oauthappdo.isRequestObjectSignatureValidationEnabled() &&
+        boolean isRequestObjectSignatureValidationEnabled = !oauthappdo.isRequestObjectSignatureValidationEnabled() &&
                 consumerAppDTO.isRequestObjectSignatureValidationEnabled();
-        if (isIdTokenEncryptionEnabled || isRequestObjectSignatureValidation) {
+        if (isIdTokenEncryptionEnabled || isRequestObjectSignatureValidationEnabled) {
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -4174,8 +4174,10 @@ private static String getSPJwksUrl(String clientId, String spTenantDomain) throw
             }
         }
         if (log.isDebugEnabled()) {
-            log.debug(String.format("Retrieved jwks uri: %s for the service provider associated with client_id: %s",
-                    jwksUri, clientId));
+            if (StringUtils.isNotBlank(jwksUri)) {
```
nilasini, line 4177: If the jwksUri is empty then we can print it as empty right? Any reason to neglect debug message if jwksUri is empty?

Maliesha: If jwksUri is empty, then it will give a debug message after the getSPJwksUrl method calls (Example - [OAuth2Util.java#L2482](https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java#L2482)). 
So checking jwksUri is not empty inside the getSPJwksUrl method to overcome the redundancy.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
```diff
@@ -262,6 +262,13 @@ public OAuthConsumerAppDTO registerAndRetrieveOAuthApplicationData(OAuthConsumer
                     }
                 }
 
+                ///Check the requirement of having a configured public certificate or JWKS in SP
+                boolean isIdTokenEncryptionEnabled = application.isIdTokenEncryptionEnabled();
+                boolean isRequestObjectSignatureValidationEnabled =
+                        application.isRequestObjectSignatureValidationEnabled();
+                handlePublicCertificateConfig(app.getOauthConsumerKey(), tenantDomain, isIdTokenEncryptionEnabled,
```
mefarazath, line 269: Shouldn't we call this method only if either isIdTokenEncryptionEnabled or isRequestObjectSignatureValidationEnabled is true

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
```diff
@@ -1819,4 +1834,42 @@ private void validateScopeExistence(String scopeName) throws IdentityOAuthAdminE
...
+     * Check the requirement of having a configured public certificate or JWKS in SP and throw an exception with a
+     * specific error message if the public certificate or JWKS in SP is not configured.
+     *
+     * @param isIdTokenEncryptionEnabled Is ID token encryption is enabled or not.
+     * @param isRequestObjectSignatureValidation Is request object signature validation is enabled or not.
+     * @param clientId Client ID of the service provider.
+     * @param tenantDomain Tenant domain of the service provider.
+     * @throws IdentityOAuthClientException
+     */
+    private void handlePublicCertificateConfig(String clientId, String tenantDomain, boolean isIdTokenEncryptionEnabled,
```
mefarazath, line 1848: Can we have a better name to reflect the function of the method?

CheckIfPublicCertConfigured or a better name :)

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
Maliesha: addressed with cc43b59fd26aeaf1931e9600fd8ac852fe1fd31d

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
Maliesha: Addressed with cef5ac44bbc380cba2702c7f898071cf525ec51f

Maliesha:
Closing since we are unable to do this due to our current approach of creating the OAuth application and the service provider application separately and merging them together.

PR #1860, opened 2022-06-22
Fix no info log when authorization request fails due to expired JWT

ThumulaPerera:
This PR introduces an info log when an authorization request fails when trying to make an authorization call with a bad JWT in the request object. For an expired JWT, the info log will mention `Expired JWT`.  

Resolves: https://github.com/wso2/product-is/issues/13934


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/RequestObjectValidatorUtil.java
```diff
@@ -149,6 +150,9 @@ public static boolean isSignatureVerified(SignedJWT signedJWT, String jwksUri) t
             try {
                 return new JWKSBasedJWTValidator().validateSignature(jwtString, jwksUri, alg, MapUtils.EMPTY_MAP);
             } catch (IdentityOAuth2Exception e) {
+                if (e.getCause() != null && e.getCause() instanceof BadJWTException) {
```
piraveena, line 153: if it is a expired JWT, then it is a client-side error. We should not log errors for client errors. What is the api response sent to the client/app when they pass the expired cert?

PR #1883, opened 2022-07-22
Skip scope validation if organization scope properties are available

wenodpathirana:
When an business application is shared to a child organization, new fragment application is getting created at the child organization. 
If an user from child organization wants to login to the application, user will be redirected to the fragment application for authentication (using OIDC). 
If OIDC has been used as inbound authentication of the (primary) application, scopes in the authorization request will be passed to the fragment application for validation / approval as well. After the validation, scopes in the Organization authenticator response will be added to the context (https://github.com/wso2-extensions/identity-organization-management/pull/75, https://github.com/wso2/carbon-identity-framework/pull/4107).

Since the scopes are handled at the fragment application, scope handling will be skipped at the primary application and scopes in authenticated results are returned in the authorize request.


mefarazath:
> ### Proposed changes in this pull request
> Skip scope validation if organization scope properties are available, since the scopes are validated at the child organization.

Can you give an example how this would work and an example of an organization scope?

wenodpathirana:
@mefarazath updated the description, scopes will be same as the scopes in the auth request, it's just the scopes are handled at the fragment application when user logged in from child child organization.

PR #1953, opened 2022-11-02
Update pom to Major Development Version

hwupathum:
Bump the version to 7.0.0 for IPK and fix dependency issues


hwupathum:
Was handled by https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1954

PR #1960, opened 2022-11-18
Check whether user claims are updated before clearing cache.

ravinduramesh:
$Subject

### Related Issues
- https://github.com/wso2/product-is/issues/15127

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/internal/OAuthApplicationMgtListener.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/internal/OAuthApplicationMgtListener.java
@@ -126,13 +126,16 @@ public class OAuthApplicationMgtListener extends AbstractApplicationMgtListener 
         return true;
     }
 
-    public boolean doPostUpdateApplication(ServiceProvider serviceProvider, String tenantDomain, String userName)
+    public boolean doPostUpdateApplication(ServiceProvider serviceProvider, String tenantDomain,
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/internal/OAuthApplicationMgtListener.java
```diff
@@ -126,13 +126,16 @@ public boolean doPostCreateApplication(ServiceProvider serviceProvider, String t
         return true;
     }
 
-    public boolean doPostUpdateApplication(ServiceProvider serviceProvider, String tenantDomain, String userName)
+    public boolean doPostUpdateApplication(ServiceProvider serviceProvider, String tenantDomain,
```
Avarjana, line 129: Since this is a public method I think it's better to go with an overloading approach.

mefarazath: +1

also can't we detect whether isClaimConfigChanged within this listener?

mefarazath: Also overloading a method just to detect whether claim config has changed doesn't seem right to me.

Also if you think about it whether isClaimConfigChanged is useful only for one particular listener or few not all listeners. So i feel it would be good to detect whether whether isClaimConfigChanged within the listener implementation logic itself

PR #2004, opened 2023-01-27
Prevent sending all scopes for SYSTEM scope requests.

ravinduramesh:
$Subject

### Related PRs
- https://github.com/wso2/product-is/pull/15437

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/Oauth2ScopeUtils.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/Oauth2ScopeUtils.java
@@ -45,7 +45,6 @@ import java.util.Map;
 import java.util.Set;
 
 import static org.wso2.carbon.identity.oauth2.Oauth2ScopeConstants.INTERNAL_SCOPE_PREFIX;
-import static org.wso2.carbon.identity.oauth2.Oauth2ScopeConstants.SYSTEM_SCOPE;
 
 /**
  * Utility functions related to OAuth 2 scopes.
@@ -338,7 +337,7 @@ public class Oauth2ScopeUtils {
             return ArrayUtils.EMPTY_STRING_ARRAY;
         }
         for (String scope : scopes) {
-            if (scope.startsWith(INTERNAL_SCOPE_PREFIX) || scope.equalsIgnoreCase(SYSTEM_SCOPE)) {
+            if (scope.startsWith(INTERNAL_SCOPE_PREFIX)) {
                 requestedScopes.add(scope);
             }
         }
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
@@ -102,7 +102,7 @@ public class JDBCPermissionBasedInternalScopeValidator {
         if (ArrayUtils.isEmpty(requestedScopes)) {
             return requestedScopes;
         }
-        List<Scope> userAllowedScopes = getUserAllowedScopes(tokReqMsgCtx.getAuthorizedUser(), requestedScopes,
+        Set<Scope> userAllowedScopes = getUserAllowedScopes(tokReqMsgCtx.getAuthorizedUser(), requestedScopes,
                 tokReqMsgCtx.getOauth2AccessTokenReqDTO().getClientId());
         String[] userAllowedScopesAsArray = getScopeNames(userAllowedScopes);
         return userAllowedScopesAsArray;
@@ -135,25 +135,25 @@ public class JDBCPermissionBasedInternalScopeValidator {
      */
     public String[] validateScope(String[] requestedScopes, AuthenticatedUser authenticatedUser, String clientId) {
 
-        List<Scope> userAllowedScopes = getUserAllowedScopes(authenticatedUser, requestedScopes, clientId);
+        Set<Scope> userAllowedScopes = getUserAllowedScopes(authenticatedUser, requestedScopes, clientId);
         String[] userAllowedScopesAsArray = getScopeNames(userAllowedScopes);
         return userAllowedScopesAsArray;
     }
 
-    private String[] getScopeNames(List<Scope> scopes) {
+    private String[] getScopeNames(Set<Scope> scopes) {
 
         return scopes.stream()
                 .map(Scope::getName).toArray(String[]::new);
     }
 
-    private List<Scope> getUserAllowedScopes(AuthenticatedUser authenticatedUser, String[] requestedScopes,
+    private Set<Scope> getUserAllowedScopes(AuthenticatedUser authenticatedUser, String[] requestedScopes,
                                              String clientId) {
 
-        List<Scope> userAllowedScopes = new ArrayList<>();
+        Set<Scope> userAllowedScopes = new HashSet<>();
 
         try {
             if (requestedScopes == null) {
-                return new ArrayList<>();
+                return new HashSet<>();
             }
             String tenantDomain = authenticatedUser.getTenantDomain();
             boolean isFederatedRoleBasedAuthzEnabled = false;
@@ -168,8 +168,7 @@ public class JDBCPermissionBasedInternalScopeValidator {
 
             Set<Scope> allScopes = getScopesOfPermissionType(tenantId);
             if (ArrayUtils.contains(requestedScopes, SYSTEM_SCOPE)) {
-                List<Scope> allScopesList = new ArrayList<>(allScopes);
-                requestedScopes = getScopeNames(allScopesList);
+                requestedScopes = getScopeNames(allScopes);
             } else {
                 // filter out the internal scopes
                 requestedScopes = Oauth2ScopeUtils.getRequestedScopes(requestedScopes);
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
@@ -241,8 +241,6 @@ public class JDBCPermissionBasedInternalScopeValidator {
             if (ArrayUtils.contains(allowedResourcesForUser, ROOT) || ArrayUtils.contains(allowedResourcesForUser,
                     PERMISSION_ROOT)) {
                 return new ArrayList<>(allScopes);
-            } else if (ArrayUtils.contains(allowedResourcesForUser, ADMIN_PERMISSION_ROOT)) {
-                return new ArrayList<>(getAdminAllowedScopes(allScopes, requestedScopes, isSystemScope));
             }
 
             for (Scope scope : allScopes) {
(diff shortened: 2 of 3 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -242,7 +242,7 @@ Note that if there is association between a federated user and local user () 'As
                     PERMISSION_ROOT)) {
                 return new ArrayList<>(allScopes);
             } else if (ArrayUtils.contains(allowedResourcesForUser, ADMIN_PERMISSION_ROOT)) {
-                return new ArrayList<>(getAdminAllowedScopes(allScopes, requestedScopes));
+                return new ArrayList<>(getAdminAllowedScopes(allScopes, requestedScopes, isSystemScope));
             }
```
mefarazath, line 246: what if we remove this whole else if block?

ravinduramesh: It solves this problem and unit tests also got passed after removing that block. But isn't it risky to simply remove a code section?

mefarazath: Not if we understand the consequences :)

We have used the if - else if block to short circuit individual permission checks when SYSTEM scope is presented.
Can we take a overall look at this method and make sure we don't have similar glitches.

For example why are we returning all the all scopes even for a super tenant user with root permissions without considering the requested scopes? (If block)
don't we need to consider the requested scopes even for a user who has root permission

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -241,8 +241,6 @@ Note that if there is association between a federated user and local user () 'As
             if (ArrayUtils.contains(allowedResourcesForUser, ROOT) || ArrayUtils.contains(allowedResourcesForUser,
```
mefarazath, line 241: Can we think of unifying the logic by rewriting this log to,

1. Have a method that evaluates requested scopes against user's permissions.
2. If a user is requesting SYSTEM scopes he is requesting all scopes in the system. We can use method introduced step 1 by passing all scopes in the system as request scopes.

Basic idea is we should not treat any user special and have short circuited paths like this :)

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -241,8 +241,6 @@ Note that if there is association between a federated user and local user () 'As
...
                     PERMISSION_ROOT)) {
                 return new ArrayList<>(allScopes);
-            } else if (ArrayUtils.contains(allowedResourcesForUser, ADMIN_PERMISSION_ROOT)) {
-                return new ArrayList<>(getAdminAllowedScopes(allScopes, requestedScopes));
             }
 
             for (Scope scope : allScopes) {
```
mefarazath, line 246: Can't we change requestedScopes to a set and improve the lookup here?

mefarazath (changes requested):
Since this is a critical part of allowing access, Can we have a code review and think about the whole method to prevent any possible glitches?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
ravinduramesh: If we assign `allScopes` to `requestedScopes` when the requested scope is SYSTEM, we will have to repeat 3 for loops that occure in line 246, 252, 256 again because `requestedScopes` is an array of strings but `allScopes` is a set of scopes.

ravinduramesh: To do this, we have to introduce another for loop that iterates over all scopes as `requestedScopes` is an array of string but `allScopes` is a set of scopes. And it will decrease the efficiency as some other places (line 109, 154) will start to iterate over all the scopes if we didn't avoid for "SYSTEM" scope.

mefarazath: requestedScopes --> can't we get this as a set?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
ravinduramesh: We can remove the if condition by doing it but is it worthy to convert a list to a set and perform set interconnection just to remove an `if` condition?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -190,6 +165,17 @@ private List<Scope> getUserAllowedScopes(AuthenticatedUser authenticatedUser, St
                 }
             }
             int tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
+
+            Set<Scope> allScopes = getScopesOfPermissionType(tenantId);
+            List<Scope> allScopesList = new ArrayList<>(allScopes);
```
ravinduramesh, line 170: put the list convertion inside the `if` block

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -149,22 +136,11 @@ public String[] validateScope(OAuthAuthzReqMessageContext authzReqMessageContext
...
-                scopesToRespond.add(scope);
-            }
-        }
-        return scopesToRespond.toArray(new String[0]);
+        String[] userAllowedScopesAsArray = getScopeNames(userAllowedScopes);
+        return userAllowedScopesAsArray;
     }
 
-    private String[] getScopes(List<Scope> scopes) {
+    private String[] getScopeNames(List<Scope> scopes) {
```
ravinduramesh, line 143: check whether we can accept scopes as a set

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -237,16 +223,8 @@ Note that if there is association between a federated user and local user () 'As
...
-            if (ArrayUtils.contains(allowedResourcesForUser, ROOT) || ArrayUtils.contains(allowedResourcesForUser,
-                    PERMISSION_ROOT)) {
-                return new ArrayList<>(allScopes);
-            } else if (ArrayUtils.contains(allowedResourcesForUser, ADMIN_PERMISSION_ROOT)) {
-                return new ArrayList<>(getAdminAllowedScopes(allScopes, requestedScopes));
-            }
-
             for (Scope scope : allScopes) {
-                if (!isSystemScope && !ArrayUtils.contains(requestedScopes, scope.getName())) {
+                if (!ArrayUtils.contains(requestedScopes, scope.getName())) {
```
ravinduramesh, line 227: Use set.contains with a set

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -256,7 +234,7 @@ Note that if there is association between a federated user and local user () 'As
                         for (String binding : scopeBinding.getBindings()) {
                             boolean isAllowed = false;
                             for (String allowedScope : allowedResourcesForUser) {
-                                if ((binding + "/").startsWith(allowedScope + "/")) {
+                                if (binding.startsWith(allowedScope)) {
```
ravinduramesh, line 237: keep "/" for binding

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -167,14 +167,14 @@ private List<Scope> getUserAllowedScopes(AuthenticatedUser authenticatedUser, St
...
-            List<Scope> allScopesList = new ArrayList<>(allScopes);
-
             if (ArrayUtils.contains(requestedScopes, SYSTEM_SCOPE)) {
+                List<Scope> allScopesList = new ArrayList<>(allScopes);
                 requestedScopes = getScopeNames(allScopesList);
             } else {
                 // filter out the internal scopes
                 requestedScopes = Oauth2ScopeUtils.getRequestedScopes(requestedScopes);
             }
+            Set<String> requestedScopesSet = new HashSet<>(Arrays.asList(requestedScopes));
```
mefarazath, line 177: Can't we build the set directly in these two methods?
We might have to modify getScopeNames() and Oauth2ScopeUtils.getRequestedScopes()

Check if its worth the effort over copying once. Otherwise this is fine.

Also how do we handle the case "SYSTEM some_random_scope"
some_random_scope is not an internal_* type scope but another scope which is registered and allowed for the app. Do they type of scopes get filtered before moving to this logic?

ravinduramesh: These functions are only for handling internal scopes. other registered scopes are handled in the parent function.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/Oauth2ScopeUtils.java
```diff
@@ -338,7 +337,7 @@ public static String[] getRequestedScopes(String[] scopes) {
             return ArrayUtils.EMPTY_STRING_ARRAY;
         }
         for (String scope : scopes) {
-            if (scope.startsWith(INTERNAL_SCOPE_PREFIX) || scope.equalsIgnoreCase(SYSTEM_SCOPE)) {
```
mefarazath, line 341: Are there any usages of this method outside of oauth component?
If not those methods might rely on this to filter the SYSTEM special scope

ravinduramesh: No. It's used only in one place and it's modified in this PR. Checked by searching over the repo.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/JDBCPermissionBasedInternalScopeValidator.java
```diff
@@ -97,26 +97,15 @@ public class JDBCPermissionBasedInternalScopeValidator {
      */
     public String[] validateScope(OAuthTokenReqMessageContext tokReqMsgCtx) {
 
-        // filter internal scopes
-        String[] requestedScopes = Oauth2ScopeUtils.getRequestedScopes(tokReqMsgCtx.getScope());
+        String[] requestedScopes = tokReqMsgCtx.getScope();
```
mefarazath, line 100: Here we aren't filtering for internal_* scope. Could this be an issue?

mefarazath: As in previously any logic beyond this would only expect internal_* scopes. Now it might get some additional scopes.
Is this the case?

ravinduramesh: No aiye. This method calls the getUserAllowedScopes() method. There we use Oauth2ScopeUtils.getRequestedScopes() to filter internal_* scopes.

ravinduramesh: Refer to modified code from line 168 to 175.

PR #2018, opened 2023-02-24
Add request binding to JWT tokens.

bhagyasakalanka:
### Purpose
Currently, when HA enable for a service provider and if each node request for a token, When the token type is opaque (default), IS keep the token in the database and return the same token for combination of application, scope, binding, user. But when it comes to JWT tokens, we cannot keep the token in the database due to the different lengths in different JWT tokens. Hence when application instances in HA request for a JWT token for the same application, scope, binding, user combination, current, IS will revoke the previous issued token and issue a new token. In that case, in HA environment, only one token will be in active state.

To solve this issue, we are adding new binding type called request binding type. When we enable this feature from configurations, we will add a new binding type to JWT tokens based on configured values. We will set a random UUID as the value of this new binding type, so when a HA environment request a token for the combination application, scope, binding, user, the binding value will be different and it will issue a new different token for each request without revoking existing token.

We can enable this feature by adding following config to the deployment.toml file.

```
[oauth.jwt.renew_token_without_revoking_existing]
enable = true
```


hasinthaindrajee (approved):

mifrazmurthaja:
Referring to the git issue: https://github.com/wso2/product-is/issues/15484

PR #2023, opened 2023-02-28
Fix: Getting HTTP - 500 Internal Server Error when invalid Basic is passed

BLasan:
**Purpose**
When a user removes one or more characters at the end of a `base64` encoded token, an `IllegalArgumentException` was throwing out and not getting a proper error message as the response. Linked issue states the scenario which we're trying to fix.

**Changes**
With the added change, it throws `InvalidApplicationClientException` with the corresponding error message and error.
Screenshot 2023-02-28 at 14 18 46

**Related Issues**
- Public Fix: https://github.com/wso2/api-manager/issues/1514

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/test/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpointTest.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/test/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpointTest.java
@@ -52,6 +52,7 @@ import org.wso2.carbon.identity.oauth.endpoint.util.TestOAuthEndpointBase;
 import org.wso2.carbon.identity.oauth.tokenprocessor.TokenPersistenceProcessor;
 import org.wso2.carbon.identity.oauth2.OAuth2Service;
 import org.wso2.carbon.identity.oauth2.ResponseHeader;
+import org.wso2.carbon.identity.oauth2.bean.OAuthClientAuthnContext;
 import org.wso2.carbon.identity.oauth2.dto.OAuth2AccessTokenReqDTO;
 import org.wso2.carbon.identity.oauth2.dto.OAuth2AccessTokenRespDTO;
 import org.wso2.carbon.identity.oauth2.model.CarbonOAuthTokenRequest;
@@ -86,7 +87,7 @@ import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertTrue;
 
 @PrepareForTest({EndpointUtil.class, IdentityDatabaseUtil.class, OAuthServerConfiguration.class,
-        CarbonOAuthTokenRequest.class, LoggerUtils.class, IdentityTenantUtil.class})
+        CarbonOAuthTokenRequest.class, LoggerUtils.class, IdentityTenantUtil.class, })
 public class OAuth2TokenEndpointTest extends TestOAuthEndpointBase {
 
     @Mock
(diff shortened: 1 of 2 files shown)


janakamarasena (approved):

mefarazath (approved):

components/org.wso2.carbon.identity.oauth.endpoint/src/test/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpointTest.java
```diff
@@ -250,11 +251,17 @@ public void testIssueAccessToken(String clientId, String authzHeader, Object par
         requestParams.put(OAuth.OAUTH_USERNAME, new String[]{USERNAME});
         requestParams.put(OAuth.OAUTH_PASSWORD, new String[]{"password"});
 
+        OAuthClientAuthnContext oAuthClientAuthnContext = new OAuthClientAuthnContext();
+        oAuthClientAuthnContext.setAuthenticated(true);
```
janakamarasena, line 255: Is directly setting the client authentication to true correct like this? I can see in the data provider there is a scenario we are sending an invalid authz header as well

components/org.wso2.carbon.identity.oauth.endpoint/src/test/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpointTest.java
```diff
@@ -347,11 +354,15 @@ public void testTokenErrorResponse(String errorCode, Object headerObj, int expec
...
```
janakamarasena, line 358: In here also i can see the data provider passing params that reflect authorization errors so is setting the client authentication like this correct?

components/org.wso2.carbon.identity.oauth.endpoint/src/test/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpointTest.java
BLasan: That won't affect the `isAuthenticated` property. Coz, the values of the `OAuthClientAuthnContext` is getting populated from this line https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/model/CarbonOAuthTokenRequest.java#L235

BLasan: Irrespective of the validness of the request, this `isAuthenticated` variable becomes `false`

janakamarasena: earlier we might not have hit that code block but now you're changing the code to always set the client authentication to true. This can lead to erroneous test results especially since the data provider is sending an authz header. Have you verified that this has no impact on the current intended test scenarios?

components/org.wso2.carbon.identity.oauth.endpoint/src/test/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpointTest.java
BLasan: Yeah, the way this units tests are written, won't affect the variable `isAuthenticated`

components/org.wso2.carbon.identity.oauth.endpoint/src/test/java/org/wso2/carbon/identity/oauth/endpoint/token/OAuth2TokenEndpointTest.java
BLasan: I have run the tests and all were passing

piraveena (approved):

PR #2029, opened 2023-03-16
Fix the issue with scopes during refresh grant

nilasini:

The fixes were done in this PR

Ideally for the refresh grant, only refresh grant-related scope validation should happen, other scope validation is not needed for the refresh grant as the scope validation is already done during the token issuing time. In this PR I have removed the extra scope validation added(role-based scope validation and etc) to refresh grant and kept only refresh grant-related scope validation.

For the refresh grant scope validation, we have implemented the following flow.
1. When using the Refresh token grant if there are no scopes provided in the request then **we don't need to do a scope validation, we should return the scopes which are used for the access token**. If there are scopes requested through the refresh token then we can divide the use cases as below. The below cases are already covered with the existing code.
             1. Request scopes that are the same as the scopes sent with the previous access token:- We should give a refresh token with the requested scopes
             2. **Request scopes that include more scopes than the one sent with a previous access token:- throw exception as it can lead to an attack.
             [According to the spec authorization server should decide](https://www.rfc-editor.org/rfc/rfc6749#section-3.3) 
             3. Request scopes that include less number of scopes than the one sent with the previous access token:-  We should only give a refresh token with the scopes which are requested through a refresh grant request.

can merge this pull request.]

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
@@ -505,7 +505,18 @@ public class AccessTokenIssuer {
             return true;
         }
         if (GrantType.REFRESH_TOKEN.toString().equals(grantType)) {
+            /*
+             In the refresh token flow, we have already completed scope validation during the initial token call and
+             issued the token with authorized scopes. Therefore, during the refresh flow we don't need to do the
+             internal scope validation again. But we need to call the grant type specific scope validation handler to
+             issue the token with only the authorized scopes.
+            */
             AuthorizationGrantHandler authzGrantHandler = authzGrantHandlers.get(grantType);
+            if (log.isDebugEnabled()) {
+                log.debug("Calling grant type specific scope validation handler for the refresh token grant and " +
+                        "omitting internal scope validation as internal scope validation already done " +
+                        "during the token issuance.");
+            }
             boolean isValidScope = authzGrantHandler.validateScope(tokReqMsgCtx);
             if (isValidScope) {
                 if (LoggerUtils.isDiagnosticLogsEnabled()) {
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
@@ -504,7 +504,22 @@ public class AccessTokenIssuer {
             }
             return true;
         }
-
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -504,7 +504,22 @@ private boolean validateScope(OAuthTokenReqMessageContext tokReqMsgCtx) throws I
             }
             return true;
         }
-
+        if (GrantType.REFRESH_TOKEN.toString().equals(grantType)) {
+            AuthorizationGrantHandler authzGrantHandler = authzGrantHandlers.get(grantType);
+            boolean isValidScope = authzGrantHandler.validateScope(tokReqMsgCtx);
```
janakamarasena, line 509: Why are we specifically calling the grant handler validate scope over here? isn't this anyway being called?

nilasini: No we are calling the grant handler specific validate scope methos and returning from here and omitting the extra validations. Because for refresh token we don't need to validate the scopes again with other scope validators since the scopes were already validated during the token issuing time.

shashimalcse (dismissed):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
mefarazath: Let's add a debug log here to indicate this

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
```diff
@@ -154,7 +150,11 @@ public boolean validateScope(OAuthTokenReqMessageContext tokReqMsgCtx)
                 return false;
             }
             List<String> grantedScopeList = Stream
-                    .concat(Arrays.stream(grantedScopes), Arrays.stream(grantedInternalScopes))
+                    .concat(grantedScopes == null
+                            ? Stream.empty()
+                            : Arrays.stream(grantedScopes), grantedInternalScopes == null
+                            ? Stream.empty()
+                            : Arrays.stream(grantedInternalScopes))
```
mefarazath, line 157: What if we do this just before creating the list?
It would be much readable IMO

mefarazath (approved):

mefarazath (approved):

PR #2057, opened 2023-04-11
Fix introspection failure for organization switched tokens

AnuradhaSK:
$Subject

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
@@ -43,6 +43,8 @@ import org.wso2.carbon.identity.oauth2.internal.OAuth2ServiceComponentHolder;
 import org.wso2.carbon.identity.oauth2.model.AccessTokenDO;
 import org.wso2.carbon.identity.oauth2.util.OAuth2Util;
 import org.wso2.carbon.utils.DiagnosticLog;
+import org.wso2.carbon.identity.organization.management.service.OrganizationManager;
+import org.wso2.carbon.identity.organization.management.service.exception.OrganizationManagementException;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
```diff
@@ -458,7 +460,23 @@ private OAuth2IntrospectionResponseDTO validateAccessToken(OAuth2TokenValidation
                         = OAuthServerConfiguration.getInstance().isCrossTenantTokenIntrospectionAllowed();
                 if (!isCrossTenantTokenIntrospectionAllowed && accessTokenDO != null &&
```
inthirakumaaran, line 461: Is enabling cross-tenant introspection necessary for the organization management feature? While cross-tenant introspection is often used for SaaS applications, it may not be relevant for B2B shared applications.WDYT?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
```diff
@@ -458,7 +460,23 @@ private OAuth2IntrospectionResponseDTO validateAccessToken(OAuth2TokenValidation
...
                         !tenantDomain.equalsIgnoreCase(accessTokenDO.getAuthzUser().getTenantDomain())) {
-                    throw new IllegalArgumentException("Invalid Access Token. ACTIVE access token is not found.");
+                    /*
+                    If isOrganizationManagementEnable is true, validate whether client's tenant relationship with
+                    authorized user tenant.
+                     */
+                    if (!OAuth2ServiceComponentHolder.getInstance().isOrganizationManagementEnabled()) {
+                        throw new IllegalArgumentException("Invalid Access Token. ACTIVE access token is not found.");
```
inthirakumaaran, line 468: Wondering about this exception type and error message, don't we have any better alternatives for the exception & for the error msg may be we can simply say `Invalid Access token`, WDYT?

AnuradhaSK: `IllegalArgumentException` is the exception used throughout introspection failures in this class. 
Therefore, shall we stick to the same exception?
Regarding the error message: I'm using the same message we had initially check line no: 461.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
AnuradhaSK: `Is enabling cross-tenant introspection necessary for the organization management feature` no
Here we are B2B application token handling under isCrossTenantTokenIntrospectionAllowed = false condition

inthirakumaaran: Sorry I missed it but my point is B2B token introspection is orthogonal to `isCrossTenantTokenIntrospectionAllowed` & it should not depend on the value of the `isCrossTenantTokenIntrospectionAllowed` at all, may if we can do the `isOrganizationManagementEnabled` logic first and if it is a success then use another variable skip the error throwing part from `isCrossTenantTokenIntrospectionAllowed` logic

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
inthirakumaaran: I can see that but IMO I don't feel those are correct as well. Regarding the error msg  `ACTIVE access token is not found` this part seems a bit misleading in this particular context, ie  the token is active but they not related to this tenant that's why its okay to say the token is invalid but not the second half which states that provided token is not active

PR #2061, opened 2023-04-20
Join IDN_OAUTH2_AUTHZ_CODE_SCOPE table for queries used when retrieving code by consumer key and user cases

mpmadhavig:
$subject

### Related issues
- https://github.com/wso2/product-is/issues/15724


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AuthorizationCodeDAOImpl.java
```diff
@@ -490,6 +491,10 @@ public List<AuthzCodeDO> getAuthorizationCodesByUserForOpenidScope(Authenticated
                     // Authorization codes that are in ACTIVE state and not expired should be removed from the cache.
                     if (OAuth2Util.getTimeToExpire(issuedTimeInMillis, validityPeriodInMillis) > 0) {
                         if (isHashDisabled) {
+                            List<String> authorizationCodeScopes =
+                                    getAuthorizationCodeScopes(connection, authzCodeId, tenantId);
+                            scope = authorizationCodeScopes.toArray(new String[0]);
```
sachiniWettasinghe, line 496: What is the reason to do another DB call here to get scopes? It should be already resolved with the initial call with the modified `GET_OPEN_ID_AUTHORIZATION_CODE_DATA_BY_AUTHZUSER ` query right?

mpmadhavig: This DB call was separately added because each rwo returned by the `GET_OPEN_ID_AUTHORIZATION_CODE_DATA_BY_AUTHZUSER` query only returns `openid` as the scope. But the DAO returned by the method using this query includes a scope field. Ideally the scope should be all the scopes requested by the user. Intension was to get all the scopes related to the code.

mpmadhavig: I have done some more digging after this question has raised. It was concluded that, the list of codes returned is only being used to clear the session data and there was no proper use of the scope field. Therefore, we can remove the above mentioned db call.

mpmadhavig: resolved with c07ede9313ed2b5b3aa50cbd359c3bab68950ec4

sachiniWettasinghe (approved):

PR #2063, opened 2023-04-23
Revoke authzCode when the user is locked

kayathiri4:
Revoke authorization code when the user is locked or deleted. So the token cannot be used to get access token after.


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AuthorizationCodeDAO.java
```diff
@@ -48,6 +48,9 @@ void deactivateAuthorizationCode(AuthzCodeDO authzCodeDO) throws
 
     Set<String> getAuthorizationCodesByUser(AuthenticatedUser authenticatedUser) throws IdentityOAuth2Exception;
 
+    List<AuthzCodeDO> getAuthorizationCodesDataByUser(AuthenticatedUser authenticatedUser)
```
sahandilshan, line 51: Shouldn't we make this default because if someone else using the same interface, they might need to do a migration after this change

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AuthorizationCodeDAOImpl.java
```diff
@@ -509,6 +509,73 @@ public List<AuthzCodeDO> getAuthorizationCodesByUserForOpenidScope(Authenticated
...
+            while (rs.next()) {
+                long validityPeriodInMillis = rs.getLong(3);
+                Timestamp timeCreated = rs.getTimestamp(2, Calendar.getInstance(TimeZone.getTimeZone(UTC)));
+                long issuedTimeInMillis = timeCreated.getTime();
+                String authorizationCode = getPersistenceProcessor().getPreprocessedAuthzCode(rs.getString(1));
+                String authzCodeId = rs.getString(4);
+                String[] scope = OAuth2Util.buildScopeArray(rs.getString(5));
+                String callbackUrl = rs.getString(6);
+                String consumerKey = rs.getString(7);
+                String idpName = rs.getString(8);
```
sahandilshan, line 553: Do we need to get all this information? This will be used to get the authz_code and revoke it, right? So getting only `authorizationCode` would be enough, right?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
```diff
@@ -514,6 +516,42 @@ public static boolean removeUserClaimsFromCache(String userName, UserStoreManage
...
+            for (AuthzCodeDO authorizationCode : authorizationCodes) {
+                OAuthCache.getInstance().clearCacheEntry(new OAuthCacheKey(
+                        OAuth2Util.buildCacheKeyStringForAuthzCode(authorizationCode.getConsumerKey(),
+                                authorizationCode.getAuthorizationCode())));
+                OAuthTokenPersistenceFactory.getInstance().getAuthorizationCodeDAO()
+                        .updateAuthorizationCodeState(authorizationCode.getAuthorizationCode(),
+                                OAuthConstants.AuthorizationCodeState.REVOKED);
+            }
+        } catch (IdentityOAuth2Exception e) {
+            throw new RuntimeException(e);
```
DMHP, line 549: This is not ideal, as it could mask the underlying issue and make it difficult to diagnose. It would be better to log the exception and re-throw a more specific exception that accurately reflects the issue.

kayathiri4: This is missed. Will update.

DMHP:
Let's write a unit/integration test for this @kayathiri4

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AuthorizationCodeDAOImpl.java
```diff
@@ -509,6 +509,73 @@ public List<AuthzCodeDO> getAuthorizationCodesByUserForOpenidScope(Authenticated
...
+    @Override
+    public List<AuthzCodeDO> getAuthorizationCodesDataByUser(AuthenticatedUser authenticatedUser) throws
+            IdentityOAuth2Exception {
+
+        if (log.isDebugEnabled()) {
+            log.debug("Retrieving authorization codes of user: " + authenticatedUser.toString());
+        }
+
+        Connection connection = IdentityDatabaseUtil.getDBConnection();
+        PreparedStatement ps = null;
```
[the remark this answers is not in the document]
kayathiri4: We need to assign it hence we are using it in finally block, so initial assignment is needed.

DMHP (approved):

PR #2020, opened 2023-02-27
Filter query params from consent page redirect url

sandushi:
### Proposed changes in this pull request

- **Filter all the redirect params from consent url by default**
- Store the filtered params in a cache to serve for data API
- Only sessionDataKeyConsent is append to the consent page url

- If we are adding this below config, the filtering is done according to the  filebased below config

 ```
 [authentication.endpoint.redirect_params]
  filter_policy = "include"
  remove_on_consume_from_api = "true"
  parameters = ["sessionDataKey"]
```

#### General

- [ ] **Is this PR explained thoroughly?** All code changes must be accounted for in the PR description.
- [ ] **Is the PR labeled correctly?**

#### Functionality

- [ ] **Are all requirements met?** Compare implemented functionality with the requirements specification.
- [ ] **Does the UI work as expected?** There should be no Javascript errors in the console; all resources should load. There should be no unexpected errors. Deliberately try to break the feature to find out if there are corner cases that are not handled.

#### Code

- [ ] **Do you fully understand the introduced changes to the code?** If not ask for clarification, it might uncover ways to solve a problem in a more elegant and efficient way.
- [ ] **Does the PR introduce any inefficient database requests?** Use the debug server to check for duplicate requests.
- [ ] **Are all necessary strings marked for translation?** All strings that are exposed to users via the UI must be [marked for translation](https://docs.djangoproject.com/en/1.10/topics/i18n/translation/).

#### Tests

- [ ] **Are there sufficient test cases?** Ensure that all components are tested individually; models, forms, and serializers should be tested in isolation even if a test for a view covers these components.
- [ ] **If this is a bug fix, are tests for the issue in place?**  There must be a test case for the bug to ensure the issue won’t regress. Make sure that the tests break without the new code to fix the issue.
- [ ] **If this is a new feature or a significant change to an existing feature?** has the manual testing spreadsheet been updated with instructions for manual testing?

#### Security

- [ ] **Confirm this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets.**
- [ ] **Are all UI and API inputs run through forms or serializers?**
- [ ] **Are all external inputs validated and sanitized appropriately?**
- [ ] **Does all branching logic have a default case?**
- [ ] **Does this solution handle outliers and edge cases gracefully?**
- [ ] **Are all external communications secured and restricted to SSL?**

#### Documentation

- [ ] **Are changes to the UI documented in the platform docs?** If this PR introduces new platform site functionality or changes existing ones, the changes should be documented.
[...]

PR #2075, opened 2023-05-12
Add the backward compatibility to the consent page url to keep query params

sandushi:
### Purpose
$subject

### Changes from the PR

- The default behaviour of the consent page url is changed from https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2020
- Previously all the query parameters are appended to the consent page url.
- Now, all the query parameters are filtered from the consent page url except sessionDataKeyConsent query param by default.
- So to keep the backward compatibility, from this PR , we introduced the below config.
- If some one needs to keep the previous behavior, needs to add the below config to deployment.toml file

```
[authentication.endpoint.consent_page_redirect_params]
allow = true
```
- Check whether the above config and keep all the query params in the consent page url, otherwise filter the all query params except sessionDataKeyConsent param.


components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/util/EndpointUtil.java
```diff
@@ -1773,4 +1776,8 @@ public static boolean isExternalConsentPageEnabledForSP(ServiceProvider serviceP
         }
         return isEnabled;
     }
+
+    public static boolean isConsentPageRedirectParamsAllowed() {
+        return FileBasedConfigurationBuilder.getInstance().isConsentPageRedirectParamsAllowed();
```
ashensw, line 1781: Add a new line after the method signature.

sandushi: Resolved by https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2075/commits/797a2d4fb1cd55a1395d0d30d6852a97b9e9edcf

ashensw (approved):

lashinijay (approved):

PR #2077, opened 2023-05-16
Fix possible NPE

AnuradhaSK:

Fix possible NPE when scopes == null and authorizedInternalScopes == null


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -797,6 +797,12 @@ private void addAuthorizedInternalScopes(OAuthTokenReqMessageContext tokReqMsgCt
                                              String[] authorizedInternalScopes) {
 
         String[] scopes = tokReqMsgCtx.getScope();
+        if (ArrayUtils.isEmpty(scopes)) {
+            scopes = new String[0];
```
ShanChathusanda93, line 801: Are we checking this empty scope array from any place and do the logic based on that?

ShanChathusanda93 (approved):

PR #2085, opened 2023-05-30
Resolve user id for organization specific flows when user ID is set as the username

Yasasr1:
In certain scenarios for organisation specific calls, the userId is set as the username value of authorized user. In such scenarios, following error will be logged and the oauth cache is not cleared properly. 

```
ERROR {org.wso2.carbon.identity.oauth.OAuthUtil} - User id cannot be found for user: <LOGGABLE-USER-NAME>
```
This PR fixes this issue.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
@@ -838,7 +838,7 @@ public final class OAuthUtil {
     }
 
     /**
-     * Resolves user id from username in scenarios where user id is set as the username in organization specific flows
+     * Resolves user id from username in scenarios where user id is set as the username in organization specific flows.
      *
      * @param authorizedUser authorized user.
      * @return userId  The user id.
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
```diff
@@ -830,4 +836,30 @@ public static String getUsername(String tenantDomain) throws IdentityApplication
         }
         return username;
     }
+
+    /**
+     * Resolves user id from username in scenarios where user id is set as the username in organization specific flows
```
kayathiri4, line 841:
```suggestion
     * Resolves user id from username in scenarios where user id is set as the username in organization specific flows.
```

kayathiri4 (dismissed):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
```diff
@@ -830,4 +836,30 @@ public static String getUsername(String tenantDomain) throws IdentityApplication
...
+     *
+     * @param authorizedUser authorized user.
+     * @return userId  The user id.
+     */
+    private static String resolveUserIdFromUsername(AuthenticatedUser authorizedUser) {
+
+        String userId = null;
+        if (StringUtils.isNotBlank(authorizedUser.getTenantDomain()) &&
```
kayathiri4, line 849: Is authorizedUser not null always?

Yasasr1: It's taken from the `accessTokenDO` so i think it should not be null

kayathiri4 (approved):

PR #2098, opened 2023-06-15
Changes applied into the code after CON_APP_KEY database constraint removed

Davids-98:

Token issuing logic in the identity server will be simplified with this change. There is a db constraint as "CON_APP_KEY" and with that logic when two requests come we issue the past record in DB for that request rather than use two records in the DB. It is not a good experience for a developer. 

So I removed the CON_APP_KEY DB constraint first. With this change identity server gives a new access token for each new token request. When there is CON_APP_KEY db constraint, the behavior of the token may differ with the time ( let's think if the expiration time of a token is one hour and expire time of the refresh token is one week. Initially, when we request a token with these conditions we get a token without any problems. when we request a token again in 30 minutes we get the same access token and the same refresh token. When we request a token again in one and half hours we get a new access token with the same refresh token. When we request a token again in two weeks we get a new access token and a new refresh token). But after the CON_APP_KEY db constraint is removed from the db layer our code logic gets very simple and efficient. We get a new token for each new token request.

Changes;
*I have gone through the code base and found what changes should I do with the new assumption after the CON_APP_KEY remove

1) Check the token-issuing logic and remove the path that checks the existing token
2) Check the refresh grant flow and add relevant changes to it
3) Check the token revocation flow and add relevant changes to it.
4) Go through the DAO layer and change relevant code scripts that were implemented with the previous assumptions


Thumimku:
Hi @Davids-98,

⚠️ **Notice:** This PR has been open for a while. To keep the repository clean and up-to-date, this PR will be **closed within the next two weeks** if there is no further activity.

Please take one of the following actions:
- **Merge the PR** if it is ready.
- **Close the PR** if it is no longer relevant.
- **Leave a comment** explaining why it should remain open and provide an update on its progress.

Your prompt attention to this matter is greatly appreciated. Thank you for your understanding and collaboration! 🙏

JeethJJ:
Closing this PR as it has been inactive for a long time. Feel free to reopen if needed.

PR #2101, opened 2023-06-16
Fix logout issue for sub-organization console by setting the userId if not exist in authenticated user object

sadilchamishka:

In this PR, a new method `setUserIdIfNotExist` is proposed to set the userId in the authenticated user if the userId is not populated in the `AuthenticatedUser` object. Normally it should be set, but when the user resides in an ancestor organization, the userId won't be able to resolve by username.

Due to that, two places were affected.

All the flows will be affected for B2B sub-organization related user logout or session extension scenarios.

1 - Building the `accessTokenDO` object when token is about to revoked upon user logout. Ideally userId should be populated, but due to above mentioned reason userId is set to `null`.

2 - When session extend request are sent using the refresh token grant type for the logged in sessions of the sub-organization level users, useID is required to cache the access token when successfully using the token. The cache key includes the userID, hence userID should be able to resolve.

In the above mentioned both cases, the tokens are based on organization switch grant type. It is due to sub-organization level tokens are issued by that grant type. More specifically, when using that grant type, the subject identifier is set in the form of `userId@tenantDomain`. That information is used to first check the existence of the user with that userID and set for authenticated user object incase the userID is missing.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
@@ -131,16 +131,19 @@ import org.wso2.carbon.identity.oauth2.token.bindings.TokenBinding;
 import org.wso2.carbon.identity.oauth2.token.handlers.grant.AuthorizationGrantHandler;
 import org.wso2.carbon.identity.openidconnect.model.Constants;
 import org.wso2.carbon.identity.openidconnect.model.RequestedClaim;
+import org.wso2.carbon.identity.organization.management.service.exception.OrganizationManagementException;
 import org.wso2.carbon.idp.mgt.IdentityProviderManagementException;
 import org.wso2.carbon.idp.mgt.IdentityProviderManager;
 import org.wso2.carbon.registry.core.Registry;
 import org.wso2.carbon.registry.core.Resource;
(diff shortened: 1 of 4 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -566,6 +569,36 @@ private static boolean isTenantActive(String tenantDomain) throws IdentityOAuth2
...
+     * @param authenticatedUser              The authenticated user object.
+     * @param authenticatedSubjectIdentifier The authenticated subject identifier.
+     */
+    public static void setUserIdIfNotExist(AuthenticatedUser authenticatedUser, String authenticatedSubjectIdentifier) {
+
+        if (StringUtils.equals(authenticatedUser.getLoggableUserId(), authenticatedUser.toFullQualifiedUsername())) {
+            if (authenticatedUser.getAuthenticatedSubjectIdentifier() == null) {
+                return;
+            }
+            String userId = authenticatedSubjectIdentifier.split("@")[0];
```
sadilchamishka, line 584: Subject identifier would be different if the subject identifier is explicitly set for an app. Hence this workaround will not help.

sadilchamishka:
Closing due to the fix is addressed by this PR - https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2103

PR #2121, opened 2023-07-26
 Fix token revocation on session expiry for SAML SLO from Federated IDP

Wathsara:
### Purpose
To fix
- https://github.com/wso2/product-is/issues/16280


nilasini:
@Wathsara any specific reason to not to get this merged?

PR #2155, opened 2023-09-04
Use Host Name As the Token iss Value

hwupathum:
Related issue https://github.com/wso2/product-is/issues/16450

Use the following config to use host name to build issuer instead of entity ID

```
[oauth]
build_issuer_with_hostname = true
```

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
@@ -3848,6 +3848,11 @@ public class OAuth2Util {
                 IDP_ENTITY_ID).getValue();
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -3847,6 +3848,20 @@ public static String getResidentIdpEntityId(String tenantDomain) throws Identity
...
+     * If enabled, hostname will be used to build the issuer of the ID token instead of entity id of the resident IDP.
+     *
+     * @return true if hostname is to be used to build the issuer of the ID token.
+     */
+    private static boolean isBuildIssuerWithHostname() {
+
+        String buildIssuerWithHostname = IdentityUtil.getProperty(OAUTH_BUILD_ISSUER_WITH_HOSTNAME);
+        if (StringUtils.isBlank(buildIssuerWithHostname)) {
+            return false;
+        }
```
ashensw, line 3861: Do we need this logic since parseBoolean is null safe ?

PR #2037, opened 2023-03-26
JARM (JWT Secured Authorization Response Mode) for OAuth 2.0 implementation

anjuchamantha:
# JARM Implementation

Fixes the issue: https://github.com/wso2/product-is/issues/15527

PR #2158, opened 2023-09-06
Preserve already existing query params in callback url when fragment response mode is used.

anjuchamantha:
Callback url can have query params in it. Currently when fragment response mode is used, those query params are also being sent as fragments. Ideally the existing query params should be there in the redirect url as query params eventhough the response mode is fragment. This PR fixes this issue https://github.com/wso2/product-is/issues/16668

This PR also removes code duplication in OAuth2AuthzEndpointTest unit tests introduced by #2037.


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/responsemode/provider/impl/FragmentResponseModeProvider.java
```diff
@@ -89,10 +88,7 @@ public String getAuthResponseRedirectUrl(AuthorizationResponseDTO authorizationR
                 queryParams.add(OAuthConstants.SCOPE + "=" + scope);
             }
 
-            redirectUrl = FrameworkUtils.appendQueryParamsStringToUrl(redirectUrl,
-                    String.join("&", queryParams));
-
-            redirectUrl = redirectUrl.replace("?", "#");
+            redirectUrl += "#" + String.join("&", queryParams);
```
anjuchamantha, line 91: Need to add unit tests & change the queryParams name to params

anjuchamantha: Fixed with later commits.

janakamarasena (approved):

PR #2161, opened 2023-09-11
Introduce PublicClientAllowed property in SupportedGrantType config

ZiyamSanthosh:
- `PublicClientAllowed` property has been introduced to define whether a grant types is allowed to be used for public client authentication.

- The current implementation let the flow bypass the condition to check if the grant type is allowed for public client authentication. It has been fixed to return false if no grant types are allowed to support public client authentication.

Related PR:
- https://github.com/wso2/carbon-identity-framework/pull/4905

Note: This PR has to be merged after merging the above mentioned PR.


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/client/authentication/PublicClientAuthenticator.java
```diff
@@ -86,7 +86,14 @@ public boolean canAuthenticate(HttpServletRequest request, Map<String, List> bod
                 getPublicClientSupportedGrantTypesList();
         List grantTypes = bodyParams.get(GRANT_TYPE);
 
-        if (!publicClientSupportedGrantTypes.isEmpty() && grantTypes != null) {
+        if (publicClientSupportedGrantTypes.isEmpty()) {
+            if (log.isDebugEnabled()) {
```
SujanSanjula96, line 90: We don't need `isDebugEnabled` check since there are no string concatenations

nilasini (dismissed):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/client/authentication/PublicClientAuthenticator.java
ZiyamSanthosh: Fixed with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2161/commits/cbe612b67ce71393be554659b387d991c78dd89e

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -2396,6 +2397,21 @@ private void parseSupportedGrantTypesConfig(OMElement oauthConfigElem) {
...
                 }
+
+                // Read the public client allowed grant types from default grant types.
+                OMElement publicClientAllowedElement = supportedGrantTypeElement
+                        .getFirstChildWithName(getQNameWithIdentityNS(ConfigElements.PUBLIC_CLIENT_ALLOWED));
+                String publicClientAllowed = null;
+                if (publicClientAllowedElement != null) {
+                    publicClientAllowed = publicClientAllowedElement.getText();
+                }
+                if (StringUtils.isNotEmpty(publicClientAllowed)) {
```
nilasini, line 2408: what will happen if we have a grant type with out the value for publicClientAllowed (if publicClientAllowed is empty)?

ZiyamSanthosh: If `PublicClientNotAllowed` property is not present, then `publicClientAllowedElement` will be null and hence, `publicClientAllowed ` will be null as well. So the loop will continue.

If the property is present and no value is provided, `publicClientAllowedElement` will not be null and `publicClientAllowed` will be an empty string (""). So the check in line 2408 will fail and hence the loop will continue.

nilasini: If it is empty or null then the relevant grant type will be added to publicClientSupportedGrantTypes or publicClientNotSupportedGrantTypes right?

ZiyamSanthosh: Yes. Sorry for the misunderstanding. Fixed with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2161/commits/049806cbd0006c7e972b2ff7b6418feeda51c4a5

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -2396,6 +2397,21 @@ private void parseSupportedGrantTypesConfig(OMElement oauthConfigElem) {
...
+                    if (Boolean.parseBoolean(publicClientAllowed)) {
+                        publicClientSupportedGrantTypes.add(grantTypeName);
+                        continue;
+                    }
+                }
+                publicClientNotSupportedGrantTypes.add(grantTypeName);
```
nilasini, line 2414:
```suggestion
             
                    if (Boolean.parseBoolean(publicClientAllowed)) {
                        publicClientSupportedGrantTypes.add(grantTypeName);
                    } else {
                         publicClientNotSupportedGrantTypes.add(grantTypeName);
                    }
```

ZiyamSanthosh: Fixed with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2161/commits/58abe255a6f640b53aad20f11ac0c0a25f8696dd

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
ZiyamSanthosh, line 2408: The validation added here was reverted to the previous implementation with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2161/commits/0bdbde689d1f94f235e62337c7ca01c87a4a747d

Explanation:
Grant types added with isPublicClientAllowed property and value set to true will be added to `publicClientSupportedGrantTypes` list and value set to false will be added to `publicClientNotSupportedGrantTypes`. All default grant types will have the property set to either the value. 
In a scenario where the property is not mentioned in the custom grant type configuration, the grant type will not be added to either lists. So, if the custom grant type is added to the old array configuration, it will get added to the `publicClientSupportedGrantTypes` list

nilasini: Can u please add a comment to the code as well so that if someone tries to modify the code they could get the clear idea.

ZiyamSanthosh: Added with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2161/commits/64f406e714a253f713f4cb583ab708cb2a7725a6

nilasini (approved):

PR #2162, opened 2023-09-11
Adding FAPI related OpenID configurations

ChinthakaJ98:
## Related Issues
https://github.com/wso2/product-is/issues/16924


components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
```diff
@@ -118,16 +118,33 @@ public OIDProviderConfigResponse buildOIDProviderConfig(OIDProviderRequest reque
...
-                            OAuthServerConfiguration.getInstance().getUserInfoJWTSignatureAlgorithm()).getName()
-            });
-        } catch (IdentityOAuth2Exception e) {
-            throw new ServerConfigurationException("Unsupported signature algorithm configured.", e);
+        List<String> supportedUserinfoSigningAlgValues = OAuthServerConfiguration.getInstance()
+                .getUserInfoJWTSignatureAlgorithms();
+        String userInfoJWTSignatureAlgorithm =  OAuthServerConfiguration.getInstance()
+                .getUserInfoJWTSignatureAlgorithm();
+        if (!supportedUserinfoSigningAlgValues.contains(userInfoJWTSignatureAlgorithm)) {
+            supportedUserinfoSigningAlgValues.add(userInfoJWTSignatureAlgorithm);
```
madurangasiriwardena, line 126: Since you are modifying the object in the OAuthServerConfiguration, there is no point in doing it again and again

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
```diff
@@ -118,16 +118,33 @@ public OIDProviderConfigResponse buildOIDProviderConfig(OIDProviderRequest reque
...
+        }
+        if (!supportedUserinfoSigningAlgValues.isEmpty()) {
```
madurangasiriwardena, line 128: Can't we do this one time and keep the structure without doing it again and again

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/OIDProviderConfigResponse.java
```diff
@@ -78,6 +78,7 @@ public class OIDProviderConfigResponse {
     private Boolean backchannelLogoutSessionSupported;
     private String[] codeChallengeMethodsSupported;
     private String deviceAuthorizationEndpoint;
+    private Boolean tlsClientCertificateBoundAccessTokens;
```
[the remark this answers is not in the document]
ChinthakaJ98: Fixed in 4b6f1443a54d3fb61af2b91bab66e7ae8b94810f

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: The object was modified in order to merge the values obtained from the existing config (UserInfoJWTSignatureAlgorithm) and the new config which supports multiple algorithms. Added a comment to the condition to give clarity to the context.

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: Modifications were moved to the OAuthServerConfiguration class in afee548fc78bfd5671aa11bc90274caee366b930

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/DiscoveryConstants.java
```diff
@@ -352,4 +352,18 @@ public class DiscoveryConstants {
      * as defined in OAuth 2.0 Device Grant [rfc8628]
      */
     public static final String DEVICE_AUTHORIZATION_ENDPOINT = "device_authorization_endpoint";
+
+    /**
+     * tls_client_certificate_bound_access_tokens
+     * OPTIONAL. Boolean value indicating server support for mutual-TLS client certificate-bound access tokens.
+     * If omitted, the default value is false.
+     */
+    public static final String TLS_CLIENT_CERTIFICATE_BOUND_ACCESS_TOKEN = "tls_client_certificate_bound_access_tokens";
```
ChinthakaJ98, line 361: Check the openid doc

ChinthakaJ98: Need to check further on this whether it should be dynamically added

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/OIDProviderConfigResponse.java
```diff
@@ -78,6 +78,7 @@ public class OIDProviderConfigResponse {
     private Boolean backchannelLogoutSessionSupported;
     private String[] codeChallengeMethodsSupported;
     private String deviceAuthorizationEndpoint;
+    private boolean isTlsClientCertificateBoundAccessTokensEnabled;
```
janakamarasena, line 81: Lets change "enabled" to "supported". let make the same change in the other relevant places
```suggestion
    private boolean isTlsClientCertificateBoundAccessTokensSupported;
```

ChinthakaJ98: Fixed in 13e64465f45bb0e715a99279ff095cc6892e9066

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
```diff
@@ -126,8 +125,12 @@ public OIDProviderConfigResponse buildOIDProviderConfig(OIDProviderRequest reque
         } catch (IdentityOAuth2Exception e) {
             throw new ServerConfigurationException("Unsupported signature algorithm configured.", e);
         }
-        providerConfig.setTokenEndpointAuthMethodsSupported(
-                OAuth2Util.getSupportedClientAuthenticationMethods().stream().toArray(String[]::new));
```
janakamarasena, line 130: the util method `getSupportedClientAuthenticationMethods` should be improved to retrieve the supported methods. 

Additionally, instead of getting these from a server config is there a way to get these values dynamically from the actually available client authentication options in the server?

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
```diff
@@ -140,6 +143,23 @@ public OIDProviderConfigResponse buildOIDProviderConfig(OIDProviderRequest reque
...
+            for (int count = 0; count < supportedTokenEndpointSigningAlgorithms.size(); count++) {
+                supportedTokenEndpointSigningAlgorithmsArray[count] =  String.valueOf(OAuth2Util
+                        .mapSignatureAlgorithmForJWSAlgorithm(supportedTokenEndpointSigningAlgorithms.get(count)));
+            }
+        } catch (IdentityOAuth2Exception e) {
+            throw new ServerConfigurationException("Unsupported signature algorithm configured.", e);
+        }
+        providerConfig.setTokenEndpointAuthSigningAlgValuesSupported(supportedTokenEndpointSigningAlgorithmsArray);
+
+        providerConfig.setTlsClientCertificateBoundAccessTokensSupported(OAuthServerConfiguration.getInstance()
```
janakamarasena, line 161: This should be added if the relevant component is in the system only. As we don't need to have a config to add that component to the system let's not rely on a config to add this for discovery

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
```diff
@@ -140,6 +143,23 @@ public OIDProviderConfigResponse buildOIDProviderConfig(OIDProviderRequest reque
...
+                .size()];
+        try {
+            for (int count = 0; count < supportedTokenEndpointSigningAlgorithms.size(); count++) {
+                supportedTokenEndpointSigningAlgorithmsArray[count] =  String.valueOf(OAuth2Util
+                        .mapSignatureAlgorithmForJWSAlgorithm(supportedTokenEndpointSigningAlgorithms.get(count)));
+            }
+        } catch (IdentityOAuth2Exception e) {
+            throw new ServerConfigurationException("Unsupported signature algorithm configured.", e);
+        }
+        providerConfig.setTokenEndpointAuthSigningAlgValuesSupported(supportedTokenEndpointSigningAlgorithmsArray);
```
janakamarasena, line 159: Improve readability

```suggestion
        List<String> mappedAlgorithms = new ArrayList<>();
try {
    for (String algorithm : supportedTokenEndpointSigningAlgorithms) {
        String convertedAlgorithm = String.valueOf(OAuth2Util
                .mapSignatureAlgorithmForJWSAlgorithm(algorithm));
        mappedAlgorithms.add(convertedAlgorithm);
    }
} catch (IdentityOAuth2Exception e) {
    throw new ServerConfigurationException("Unsupported signature algorithm configured.", e);
}
providerConfig.setTokenEndpointAuthSigningAlgValuesSupported(mappedAlgorithms.toArray(new String[0]));
```

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -3530,6 +3553,67 @@ public void setGlobalRbacScopeIssuerEnabled(boolean globalRbacScopeIssuerEnabled
...
+            }
+        }
+    }
+
+    /**
+     * Parse supported signing algorithms and add them to the supportedTokenEndpointSigningAlgorithms list.
+     *
+     * @param algorithms OMElement of supported algorithms.
+     */
+    private void parseSupportedTokenEndpointSigningAlgorithms(OMElement algorithms) {
```
janakamarasena, line 3599: Instead of doing the mapped algorithm check in the ProviderConfigBuilder we can do it here itself and maintain the list. This way we can detect any failures early on as well.

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: We can get the authenticators dynamically, but the issue we come across then is that we do not have a method of mapping the authenticator class with the authenticator referring name. As an example for the PrivateKeyJWT authenticator we use the referring term as 'private_key_jwt' which cannot be mapped with the authenticator. But there is a way to overcome this which is by adding a constant to each authenticator with the referring name and then use that constant to do the mapping. Shall we go ahead with that approach? WDYT?

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: As the MTLS authenticator already supports it, shall we hardcode this as true without using a config? Please note that in the FAPI conformance test suite this is being checked from the well-known.

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: Fixed in a0e00c2b1fae6d3e771343b8dcc8840d1787da91

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfigurationTest.java
```diff
@@ -430,6 +431,30 @@ public void testIsAccessTokenPartitioningEnabled() throws Exception {
...
+    public void testGetSupportedTokenEndpointAuthMethods() {
+
+        List<String> supportedTokenEndpointAuthMethods = OAuthServerConfiguration.getInstance()
+                .getSupportedTokenEndpointAuthMethods();
+        Assert.assertTrue(supportedTokenEndpointAuthMethods.contains("client_secret_basic"));
+        Assert.assertTrue(supportedTokenEndpointAuthMethods.size() == 4);
+    }
+
+    @Test
+    public void testGetSupportedTokenEndpointSigningAlgorithms() {
```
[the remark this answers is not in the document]
ChinthakaJ98: Fixed in f1572cfb34dfd0d832d68219856bc7e1d13399ea

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: If not we can write a separate method from which we could do that mapping based on the class name of the authenticator. But the issue we come across then is that if a custom authenticator is added, it will be skipped as the mappings will be hardcoded in this said method.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
ChinthakaJ98: Fixed in a0e00c2b1fae6d3e771343b8dcc8840d1787da91

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
janakamarasena: Shall we add a new method to the interface so that implementing authenticators should define the this reference name.

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
janakamarasena: can we check if this component is available in the system and set this to true.
If we go in that lets add that checking logic to a method inside the OAuth2Util

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: Yeah. Thats better. I will create separate PRs to add it to the interface and to update the existing authenticators.

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: In that case what we need to check is whether the MTLS authenticator is present because the logic is in there. So will add that check to a separate method in OAuth2Util.

janakamarasena: yes sounds good

ChinthakaJ98: Fixed in ca2fd496047be88c3a8d0c0721f3b7ea411a3209

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -3568,6 +3582,40 @@ public void setGlobalRbacScopeIssuerEnabled(boolean globalRbacScopeIssuerEnabled
...
+            throws ServerConfigurationException {
+
+        if (algorithms == null) {
+            return;
+        }
+
+        Iterator iterator = algorithms.getChildrenWithLocalName(
+                ConfigElements.SUPPORTED_TOKEN_ENDPOINT_SIGNING_ALG);
+        if (iterator != null) {
+            for (; iterator.hasNext(); ) {
```
janakamarasena, line 3605: lets replace with a while loop. improves readability

```suggestion
            while (iterator.hasNext()) {
```

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -4960,4 +4963,38 @@ public static String[] extractCredentialsFromAuthzHeader(HttpServletRequest requ
 
         return OAuthUtils.decodeClientAuthenticationHeader(authorizationHeader);
     }
+
+    /**
+     * Check whether TLS client certificate bound access tokens are supported by the server.
+     *
+     * @return     Whether TLS client certificate bound access tokens are supported.
+     */
+    public static boolean isTlsClientCertificateBoundAccessTokensSupported() {
```
janakamarasena, line 4972: In the last discussion, we discussed this is not bound to an authenticator and also this should be a binding type. Lets change this with that direction.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -4960,4 +4963,38 @@ public static String[] extractCredentialsFromAuthzHeader(HttpServletRequest requ
...
+
+    /**
+     * Retrieve the list of client authentication methods supported by the server.
+     *
+     * @return     Client authentication methods supported by the server.
+     */
+    public static String[] getSupportedClientAuthMethods() {
+
+        List<OAuthClientAuthenticator> clientAuthenticators = OAuth2ServiceComponentHolder.getAuthenticationHandlers();
+        List<String> supportedClientAuthMethods = new ArrayList<>();
```
janakamarasena, line 4991: use a hashset to make sure that no duplicate values gets added.

ChinthakaJ98: Fixed in 3c54ef90dce0b0bb1fc707a493b621553251eba8

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
ChinthakaJ98: Removed TLS certificate bound access token related changes in 3c54ef90dce0b0bb1fc707a493b621553251eba8. Will add that configuration after moving towards the discussed direction.

components/org.wso2.carbon.identity.discovery/src/main/java/org/wso2/carbon/identity/discovery/builders/ProviderConfigBuilder.java
ChinthakaJ98: This concern was addressed from the following PRs.
[wso2-extensions/identity-inbound-auth-oauthh#2188](https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2188)
https://github.com/wso2-extensions/identity-oauth-addons/pull/110

janakamarasena (approved):

PR #2167, opened 2023-09-18
Fix JWT token cache clearing issue

lashinijay:
### Purpose

>  Cache is not being properly cleared when a user attempts to obtain a token using the password grant after having used the refresh grant. This issue arises because when dealing with JWT tokens. With this PR aforementioned issue will be resolved.

### Public Git Issue

- https://github.com/wso2/product-is/issues/16634

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
@@ -1017,8 +1017,7 @@ public abstract class AbstractAuthorizationGrantHandler implements Authorization
         return OAuthServerConfiguration.getInstance().isTokenRenewalPerRequestEnabled();
     }
 
-    private void clearExistingTokenFromCache(OAuthTokenReqMessageContext tokenMsgCtx, AccessTokenDO existingTokenBean)
-            throws IdentityOAuth2Exception {
+    private void clearExistingTokenFromCache(OAuthTokenReqMessageContext tokenMsgCtx, AccessTokenDO existingTokenBean) {
 
         if (cacheEnabled) {
             String tokenBindingReference = getTokenBindingReference(tokenMsgCtx);
@@ -1035,8 +1034,11 @@ public abstract class AbstractAuthorizationGrantHandler implements Authorization
                     OAuthUtil.clearOAuthCache(tokenAlias);
                 }
             } catch (OAuthSystemException e) {
-                throw new IdentityOAuth2Exception("Error while clearing the cache for the access token " +
-                        existingTokenBean.getAccessToken(), e);
+                String errorMessage = "Error while clearing cache for the access token";
+                if (IdentityUtil.isTokenLoggable(IdentityConstants.IdentityTokens.ACCESS_TOKEN)) {
+                    errorMessage += " : " + existingTokenBean.getAccessToken();
+                }
+                log.error(errorMessage, e);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -1026,6 +1029,19 @@ private void clearExistingTokenFromCache(OAuthTokenReqMessageContext tokenMsgCtx
                     OAuth2Util.buildScopeString(existingTokenBean.getScope()));
             OAuthUtil.clearOAuthCache(existingTokenBean.getConsumerKey(), existingTokenBean.getAuthzUser());
             OAuthUtil.clearOAuthCache(existingTokenBean);
+            if (REFRESH_TOKEN_GRANT_TYPE.equals(existingTokenBean.getGrantType())) {
+                try {
+                    OauthTokenIssuer tokenIssuer = OAuth2Util.getOAuthTokenIssuerForOAuthApp(
```
janakamarasena, line 1034: The used method can result in a db call. This needed info should already be available through the methods that's called previously. lets find that info from those places without needing a db call.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -1026,6 +1029,19 @@ private void clearExistingTokenFromCache(OAuthTokenReqMessageContext tokenMsgCtx
                     OAuth2Util.buildScopeString(existingTokenBean.getScope()));
             OAuthUtil.clearOAuthCache(existingTokenBean.getConsumerKey(), existingTokenBean.getAuthzUser());
             OAuthUtil.clearOAuthCache(existingTokenBean);
+            if (REFRESH_TOKEN_GRANT_TYPE.equals(existingTokenBean.getGrantType())) {
```
janakamarasena, line 1032: We should not have grant-specific logic at this level.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -1026,6 +1028,15 @@ private void clearExistingTokenFromCache(OAuthTokenReqMessageContext tokenMsgCtx
                     OAuth2Util.buildScopeString(existingTokenBean.getScope()));
             OAuthUtil.clearOAuthCache(existingTokenBean.getConsumerKey(), existingTokenBean.getAuthzUser());
             OAuthUtil.clearOAuthCache(existingTokenBean);
+            try {
+                if (oauthTokenIssuer != null && OAuth2TokenUtil.isJWT(existingTokenBean.getAccessToken())) {
+                    String tokenAlias = oauthTokenIssuer.getAccessTokenHash(existingTokenBean.getAccessToken());
+                    OAuthUtil.clearOAuthCache(tokenAlias);
+                }
+            } catch (OAuthSystemException e) {
+                log.error("Error while clearing cache for the access token : " +
```
nilasini, line 1037: better to throw the exception and handle it in upper lelve. You can catch and throw IdentityOAuth2Exception as already IdentityOAuth2Exception is thrown from the caller.

nilasini (dismissed):

nilasini (approved):

nilasini (approved):

PR #2214, opened 2023-10-24
Remove scope param for unnecessary flows in authorize call

anjuchamantha:

This PR removes the scope param from the /authorize response in the places it is not necessary.
It is identified as an unnecessary param to the redirected callbackurl (authorize response) in a test from the FAPI Conformance Test suite [1]. This behaviour is not in wso2is-6.1 and has been introduced with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2037.

The issue was, scope param is set to the AuthorizationResponseDTO at the beginning, making it returned every time. With this PR, scope is set in the place it is needed (in `setScopes()` method), same as in wso2is-6.1.

This PR fixes the issue: https://github.com/wso2/product-is/issues/16749

References:
[1] [FAPI Conformance Test Suite](https://openid.net/certification/certification-fapi_op_testing/)

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
+++ b/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java
@@ -538,7 +538,6 @@ public class OAuth2AuthzEndpoint {
         authorizationResponseDTO.setState(oauth2Params.getState());
         authorizationResponseDTO.setResponseMode(oauth2Params.getResponseMode());
         authorizationResponseDTO.setResponseType(oauth2Params.getResponseType());
-        authorizationResponseDTO.getSuccessResponseDTO().setScope(oauth2Params.getScopes());
 
         return authorizationResponseDTO;
     }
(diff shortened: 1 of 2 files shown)


chamathns:
Isn't the existing behavior aligning with the oauth spec?
https://datatracker.ietf.org/doc/html/rfc6749#section-5.1

```
   scope
         OPTIONAL, if identical to the scope requested by the client;
         otherwise, REQUIRED.  The scope of the access token as
         described by Section 3.3.
```

Ideally, we should only send back the mismatching scopes, which is a must. But sending back the user requested scopes as they are also aligns with the spec iinm.

anjuchamantha:
> Isn't the existing behavior aligning with the oauth spec? https://datatracker.ietf.org/doc/html/rfc6749#section-5.1
> 
> ```
>    scope
>          OPTIONAL, if identical to the scope requested by the client;
>          otherwise, REQUIRED.  The scope of the access token as
>          described by Section 3.3.
> ```
> 
> Ideally, we should only send back the mismatching scopes, which is a must. But sending back the user requested scopes is fine since it also aligns with the spec iinm.

Here in spec it only mention this for issuing an Access Token right? Existing behavior aligns in access token issuing flow. It is not changed with this PR. The issue is scope is returned eachtime the /authorize call is made (it is not mentioned in the spec for other flows), which was not the case before JARM.

chamathns (approved):

PR #2229, opened 2023-10-28
Fix for federated user scope validation

ThaminduR:

$subject


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/validationhandler/impl/RoleBasedScopeValidationHandler.java
```diff
@@ -111,7 +112,9 @@ private List<String> getRoleIdsAssociatedWithApp(String appId, String tenantDoma
 
         try {
             return OAuth2ServiceComponentHolder.getApplicationMgtService()
-                    .getAssociatedRolesOfApplication(appId, tenantDomain).stream().map(RoleV2::getId)
+                    .getAssociatedRolesOfApplication(appId,
+                            PrivilegedCarbonContext.getThreadLocalCarbonContext().getTenantDomain())
```
sadilchamishka, line 116: Why we need the tenant domain from the context when a tenant domain is given. Any difference?

ThaminduR: Given tenant domain is where the user resides in, in this case would be the organization. The role doesn't exist there.

PR #2234, opened 2023-11-01
Treat APPLICATION type token requests with organization_switch grant same as CC grant at the scope validator level

sadilchamishka:

$subject

### Related Issues
- https://github.com/wso2/product-is/issues/17532


ThaminduR (approved):
LGTM

PR #2304, opened 2023-12-14
Add ClientInstanceTokenBinding type

VivekVinushanth:
This PR adds new binding type where the client instance can send an identifier of its preference and request the IS to bind itself to the token. This provides the capability for grant types like password or token exchange where session/cookie based binding types cannot be used and also ensures each client instance gets token so that eventually the client has multiple active access token. 

Fixes: https://github.com/wso2/product-is/issues/18600

can merge this pull request.]

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
@@ -1066,7 +1066,7 @@ public class AccessTokenIssuer {
                 throw new IdentityOAuth2ClientException(OAuth2ErrorCodes.INVALID_REQUEST,
                         "TLS certificate not found in the request.");
             }
-            if (OAuth2Constants.TokenBinderType.CLIENT_INSTANCE.equals(tokenBinder.getBindingType())) {
+            if (OAuth2Constants.TokenBinderType.CLIENT_REQUEST.equals(tokenBinder.getBindingType())) {
                 // Treat as 'None' token binding requests.
                 tokReqMsgCtx.setTokenBinding(null);
                 return;
@@ -1076,7 +1076,7 @@ public class AccessTokenIssuer {
                             .getBindingType());
         }
 
-        if (OAuth2Constants.TokenBinderType.CLIENT_INSTANCE.equals(tokenBinder.getBindingType()) &&
+        if (OAuth2Constants.TokenBinderType.CLIENT_REQUEST.equals(tokenBinder.getBindingType()) &&
                 tokenBindingValueOptional.get().length() >= MAX_ALLOWED_LENGTH) {
                 throw new IdentityOAuth2ClientException(OAuth2ErrorCodes.INVALID_REQUEST,
                         "Token binding reference length exceeds limit");
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
@@ -1079,7 +1079,8 @@ public class AccessTokenIssuer {
         if (OAuth2Constants.TokenBinderType.CLIENT_INSTANCE.equals(tokenBinder.getBindingType()) &&
                 tokenBindingValueOptional.get().length() >= MAX_ALLOWED_LENGTH) {
                 throw new IdentityOAuth2ClientException(OAuth2ErrorCodes.INVALID_REQUEST,
-                        "Token binding reference length exceeds limit for the token binder: " + tokenBinder.getBindingType());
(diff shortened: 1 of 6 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -1065,11 +1066,23 @@ private void handleTokenBinding(OAuth2AccessTokenReqDTO tokenReqDTO, String gran
                 throw new IdentityOAuth2ClientException(OAuth2ErrorCodes.INVALID_REQUEST,
                         "TLS certificate not found in the request.");
             }
+            if (OAuth2Constants.TokenBinderType.CLIENT_INSTANCE.equals(tokenBinder.getBindingType())) {
```
senthalan, line 1069: Having specific checks for implementation in the core is not a good practice. What is the reason for these checks?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -1065,11 +1066,23 @@ private void handleTokenBinding(OAuth2AccessTokenReqDTO tokenReqDTO, String gran
...
+                tokReqMsgCtx.setTokenBinding(null);
+                return;
+            }
             throw new IdentityOAuth2Exception(
                     "Token binding reference cannot be retrieved form the token binder: " + tokenBinder
                             .getBindingType());
         }
 
+        if (OAuth2Constants.TokenBinderType.CLIENT_INSTANCE.equals(tokenBinder.getBindingType()) &&
+                tokenBindingValueOptional.get().length() >= MAX_ALLOWED_LENGTH) {
```
senthalan, line 1080: can't we do this validation within the ClientInstanceTokenBinder itself ?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/bindings/impl/ClientInstanceTokenBinder.java
```diff
@@ -0,0 +1,116 @@
...
+public class ClientInstanceTokenBinder extends AbstractTokenBinder {
+
+    private static final String CLIENT_INSTANCE_REF = "clientInstanceRef";
+
+    @Override
+    public Optional<String> getTokenBindingValue(OAuth2AccessTokenReqDTO oAuth2AccessTokenReqDTO) {
+
+        RequestParameter[] parameters = oAuth2AccessTokenReqDTO.getRequestParameters();
+        for (RequestParameter parameter : parameters) {
+            if (CLIENT_INSTANCE_REF.equals(parameter.getKey())
```
senthalan, line 49: can we verify how easy is to send this parameter in the token request using the SDk ? does SDKs allow to add extra parameter in the token request ?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
VivekVinushanth: We can but we are unable to throw a client exception from the binder itself.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/bindings/impl/ClientInstanceTokenBinder.java
VivekVinushanth: AFAIK, our SDKs do not support sending additional parameters. But wondering back-channel grant types like password/token-exchange require an SDK?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
VivekVinushanth: If the client instance has not sent any value, we treat as no binding type (as the binding is set to null by default)

senthalan (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/bindings/impl/ClientRequestTokenBinder.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright (c) 2022, WSO2 Inc. (http://www.wso2.com).
```
kayathiri4, line 2: Shall we update License header

PR #2322, opened 2024-01-17
Add api based logout support

janakamarasena:
This PR provides the support for API based logout. The logout request can be initiated as a normal OIDC logout. If the app doesn't send cookies then the id_token is required to perform the logout else the request will fail. Additionally, SLO is not supported when cookies are not sent.

To revoke the tokens during logout use sso-session binding with token revocation enabling config.

Resolves https://github.com/wso2/product-is/issues/18004
related to https://github.com/wso2/product-is/issues/15684

### DIFF
--- a/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/model/APIError.java
+++ b/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/model/APIError.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2024, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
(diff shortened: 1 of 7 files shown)


components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/model/APIError.java
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (c) 2024, WSO2 LLC. (https://www.wso2.com) All Rights Reserved.
```
madurangasiriwardena, line 2:
```suggestion
 * Copyright (c) 2024, WSO2 LLC. (https://www.wso2.com)
```

janakamarasena: fixed

madurangasiriwardena (approved):

PR #2361, opened 2024-02-09
Handle server errors which occurs from axiom while decoding

chashikajw:
### Purpose
This will handle errors which occurs while decoding the auth header from axiom library

### Fixes
-  https://github.com/wso2/api-manager/issues/2460


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/client/authentication/BasicAuthClientAuthenticator.java
```diff
@@ -33,7 +32,9 @@
 import org.wso2.carbon.identity.oauth2.model.ClientAuthenticationMethodModel;
 import org.wso2.carbon.identity.oauth2.util.OAuth2Util;
 
+
```
ThaminduDilshan, line 35:
```suggestion
```

ThaminduDilshan (approved):

PR #2364, opened 2024-02-13
Add Backward Compatibility for legacy permission access for user based auth

SujanSanjula96:
Issue - https://github.com/wso2/product-is/issues/19467

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/DefaultOAuth2ScopeValidator.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/DefaultOAuth2ScopeValidator.java
@@ -18,6 +18,7 @@
 
 package org.wso2.carbon.identity.oauth2.validators;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.ArrayUtils;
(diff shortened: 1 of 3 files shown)


shashimalcse (dismissed):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/DefaultOAuth2ScopeValidator.java
```diff
@@ -177,8 +178,11 @@ private List<String> getAuthorizedScopes(List<String> requestedScopes, Authentic
         if (requestedScopes.isEmpty()) {
             return approvedScopes;
         }
-        List<String> internalScopes = getInternalScopes(tenantDomain);
+        List<String> internalScopes = new ArrayList<>();
         if (requestedScopes.contains(SYSTEM_SCOPE)) {
+            if (CollectionUtils.isEmpty(internalScopes)) {
```
ThaminduR, line 183: do we need the empty check here?

SujanSanjula96: added this as an additional check if we resolve this in outer block when doing improvements so that performance will be not impacted

shashimalcse (approved):

PR #2370, opened 2024-02-14
Add improvements to return response for additional DCR metadata 

chamathns:
$subject


chamathns:
Included with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2371

PR #2378, opened 2024-02-21
Fix NPE when deleting an authorized API

ThaminduR:

$subject

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/IdentityOauthEventHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/IdentityOauthEventHandler.java
@@ -230,11 +230,11 @@ public class IdentityOauthEventHandler extends AbstractEventHandler {
                 AuthorizedAPI authorizedAPI = OAuthComponentServiceHolder.getInstance()
                         .getAuthorizedAPIManagementService()
                         .getAuthorizedAPI(appId, apiId, tenantDomain);
-                List<String> removedScopes = new ArrayList<>();
-                removedScopes.addAll(authorizedAPI.getScopes().stream()
-                                .map(Scope::getName).filter(scope ->
-                        !removedScopes.contains(scope)).collect(Collectors.toList()));
-
+                if (authorizedAPI.getScopes() == null) {
+                    return;
+                }
+                List<String> removedScopes = authorizedAPI.getScopes().stream()
+                        .map(Scope::getName).collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/IdentityOauthEventHandler.java
```diff
@@ -230,11 +230,11 @@ public void handleEvent(Event event) throws IdentityEventException {
                 AuthorizedAPI authorizedAPI = OAuthComponentServiceHolder.getInstance()
                         .getAuthorizedAPIManagementService()
                         .getAuthorizedAPI(appId, apiId, tenantDomain);
-                List<String> removedScopes = new ArrayList<>();
-                removedScopes.addAll(authorizedAPI.getScopes().stream()
-                                .map(Scope::getName).filter(scope ->
-                        !removedScopes.contains(scope)).collect(Collectors.toList()));
-
+                if (authorizedAPI.getScopes() == null) {
+                    return;
```
Thumimku, line 234: can we use isempty

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/IdentityOauthEventHandler.java
```diff
@@ -230,11 +230,11 @@ public void handleEvent(Event event) throws IdentityEventException {
...
+                }
+                List<String> removedScopes = authorizedAPI.getScopes().stream()
```
Thumimku, line 236: Will this handle duplicate scopes, don't we need to check removedScopes.contains(scope) before adding or we can make it as `set`

SujanSanjula96 (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/IdentityOauthEventHandler.java
ThaminduR: this object itself can be null

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/IdentityOauthEventHandler.java
ThaminduR: there can't be any duplicated scopes coming in the authorized scopes list

PR #2406, opened 2024-03-13
Select DPoP Token Validator for DPoP Type JWT Access Tokens

PasinduLakshan99:
## Purpose
When selecting token validators for access tokens, both opaque and JWT tokens with the identifier `DPoP` should utilize the dpop token validator. However, due to a bug in the existing implementation, when the token type is JWT and its identifier is `DPoP`, the token is mistakenly validated using the bearer JWT token validator instead of the intended dpop token validator.

## Fix 
an additional condition has been introduced to ensure that the bearer JWT token validator is chosen only when the token identifier is not `DPoP`. Consequently, this modification ensures that the dpop token validator is appropriately selected for both JWT and opaque tokens with the `DPoP` token identifier.

## Related Issues
- https://github.com/wso2/product-is/issues/16606


Thumimku (approved):
LGTM

PR #2417, opened 2024-03-27
[6.13.x] Improve cache clearing for AppInfoCache and AuthorizationGrantCache

RakhithaRR:

Fixes an issue where the AppInfoCache was not cleared upon an update to the SP.

Related issue: https://github.com/wso2/api-manager/issues/2545

Back ported from https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2416

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/internal/OAuthApplicationMgtListener.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/internal/OAuthApplicationMgtListener.java
@@ -142,10 +142,8 @@ public class OAuthApplicationMgtListener extends AbstractApplicationMgtListener 
         revokeAccessTokensWhenSaaSDisabled(serviceProvider, tenantDomain);
         addClientSecret(serviceProvider);
         updateAuthApplication(serviceProvider);
+        removeEntriesFromCache(serviceProvider, tenantDomain);
 
-        if (threadLocalForClaimConfigUpdates.get()) {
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/internal/OAuthApplicationMgtListener.java
```diff
@@ -463,11 +461,11 @@ private void removeEntriesFromCache(Set<String> consumerKeys) throws IdentityOAu
                 OAuthCache.getInstance().clearCacheEntry(new OAuthCacheKey(oauthKey));
             }
 
-            if (isNotEmpty(accessTokenDOSet)) {
+            if (isNotEmpty(accessTokenDOSet) && threadLocalForClaimConfigUpdates.get()) {
```
DMHP, line 464: @JeethJJ should we conditionally clear the cache against the tokens as well ? I thought it is only the authorization code.

JeethJJ: @DMHP will confirm if we are storing anything against the token that should not be cleared and provide an update here.

JeethJJ: @DMHP We are storing certain claims of a federated user against the token in the cache as well. Hence this validation is required.

DMHP (approved):

DMHP:
@JeethJJ Please send the same to the master branch

PR #2453, opened 2024-05-10
Remove root org tokens when revoking tokens during application un-share

Yasasr1:
### Purpose

When un-sharing sub organizations, root organization tokens should not be checked.


sadilchamishka (dismissed):

sadilchamishka (approved):

Yasasr1:
pr builder failed due to the below unrelated intermittent failure
```
SelfRegisterTestCase.testSelfRegisterWithInvalidPassword:79 Self register user password invalid expected:<400> but was:<409>
```

PR #2459, opened 2024-05-16
Modify scope validation for impersonated access token

Thumimku:
Public Issue: https://github.com/wso2/product-is/issues/20066

## Purpose
Like authorization code, scope validation will happen for impersonation in subject token request flow. Hence when exchanging the token we don't need to validate the scope again for token exchange grant type. 

In the impersonation flow, we have already completed scope validation during the /authorize call and issued a subject token with the authorized scopes. During the token flow we only consider the scopes bound to the issued subject token and simply ignore any 'scope' parameter sent in the subsequent token request. Therefore, it does not make sense to go through scope validation again as there won't be any new scopes to validate.

 But we should validate scope for org switch grant because the request can have scope as body param.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
@@ -629,6 +629,20 @@ public class AccessTokenIssuer {
 
         OAuth2AccessTokenReqDTO tokenReqDTO = tokReqMsgCtx.getOauth2AccessTokenReqDTO();
         String grantType = tokenReqDTO.getGrantType();
+        if (tokReqMsgCtx.isImpersonationRequest() && OAuthConstants.GrantTypes.TOKEN_EXCHANGE.equals(grantType)) {
+            /*
+             In the impersonation flow, we have already completed scope validation during the /authorize call and
+             issued a subject token with the authorized scopes. During the token flow we only consider the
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -629,6 +629,20 @@ private boolean validateScope(OAuthTokenReqMessageContext tokReqMsgCtx) throws I
 
         OAuth2AccessTokenReqDTO tokenReqDTO = tokReqMsgCtx.getOauth2AccessTokenReqDTO();
         String grantType = tokenReqDTO.getGrantType();
+        if (tokReqMsgCtx.isImpersonationRequest() && OAuthConstants.GrantTypes.TOKEN_EXCHANGE.equals(grantType)) {
+            /*
+             In the impersonation flow, we have already completed scope validation during the /authorize call and
```
DMHP, line 634: What if we filter out the intersection of the scopes that passed through the subject token and the token request?

Thumimku: If the scope body param present in the token exchange request, we should get the intersection. This will be handled in https://github.com/wso2-extensions/identity-oauth2-grant-token-exchange/pull/26/files#diff-2d5991ab029cd499dc04f0cb2adeb35f782d03473b7c86d9392cc7e98d78ba9bR241

DMHP (approved):

PR #2462, opened 2024-05-16
Impersonation sub token issuer

Thumimku:
Public Issue: https://github.com/wso2/product-is/issues/20066

## Purpose
Like Access Token issuer, we need to have a class to govern the issueance of subject token.
We need to introduce impersonation validators which let's us to know whether its validated to issue subject token.

Fix is developed is a way that it can add custom validators if its exposed as osgi service, The validation execution will be on priority order.

- [x] Unit Tests Covered [Link](https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2462/files#diff-1e0a23687de3a43d683072dbd2dbb29133b9467877c1db7ea66d479a55845d57)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/impersonation/validators/ImpersonationValidator.java
```diff
@@ -0,0 +1,57 @@
...
+     * Validates an impersonation request based on the provided impersonation context and request DTO.
+     *
+     * @param impersonationContext      the impersonation context containing information about the validation process
+     * @param impersonationRequestDTO   the impersonation request DTO containing information about the request
+     * @return an {@code ImpersonationContext} object representing the validation context,
+     *         including validation status and any validation failure details
+     * @throws IdentityOAuth2Exception if an error occurs during impersonation request validation
+     */
+    public ImpersonationContext validateImpersonation(ImpersonationContext impersonationContext,
+                                                      ImpersonationRequestDTO impersonationRequestDTO)
```
Thumimku, line 55: remove this DTO

DMHP (approved):

PR #2473, opened 2024-05-25
Improve get authenticated user by authz code when a cache miss for organization SSO user

sadilchamishka:

The authenticate user object is retrieved by the authz code from the cache. But there can be situations where cache miss happens, due to server restart or mainly due to token request send for a different node of IS cluster(this case is very frequent).
In such case of cache miss, the authenticated user is resolved from the data persisted in the `IDN_OAUTH2_AUTHORIZATION_CODE` table. But there is no cloumn for keep the user accessed organization, hence that information get missed.

As a workaround, in this PR, the authenticated user's attributes are fetched to find the user accessed organization.

### Depends on
- https://github.com/wso2/carbon-identity-framework/pull/5719

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
@@ -23,12 +23,17 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.wso2.carbon.identity.application.authentication.framework.exception.UserIdNotFoundException;
+import org.wso2.carbon.identity.application.authentication.framework.model.AuthenticatedUser;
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
```diff
@@ -638,4 +647,26 @@ private void clearAccessTokenOAuthCache(AccessTokenDO accessTokenDO) {
             }
         }
     }
+
+    private void resolveUserResidentOrgForOrganizationSSOUsers(AuthenticatedUser authenticatedUser, String authzCode) {
+
+        if (authenticatedUser.isFederatedUser() && FrameworkConstants.ORGANIZATION_LOGIN_IDP_NAME
+                .equals(authenticatedUser.getFederatedIdPName())) {
+            String userResideOrganization = resolveUserResidentOrganization(AuthorizationGrantCache.getInstance()
+                    .getValueFromCacheByCode(new AuthorizationGrantCacheKey(authzCode)).getUserAttributes());
```
AnuradhaSK, line 656: what if this cache is missed?

AnuradhaSK (approved):

PR #2476, opened 2024-05-28
Handle token issuing for invalid org related authz codes

AnuradhaSK:
$Subject


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -141,6 +142,14 @@ public OAuth2AccessTokenRespDTO issue(OAuthTokenReqMessageContext tokReqMsgCtx)
         String tokenBindingReference = getTokenBindingReference(tokReqMsgCtx);
         String authorizedOrganization = getAuthorizedOrganization(tokReqMsgCtx);
 
+        // If the authorizedOrganization is deactivated or not available, return an error.
+        if (!(OAuthConstants.AuthorizedOrganization.NONE.equals(authorizedOrganization) ||
```
sadilchamishka, line 146: Can we get "NONE" for the authorizedOrganization. It should be null or any value. If it is coming NONE, we may have to check and fix that (as it might be a mistake).

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -1424,6 +1426,25 @@ public static String getIDTokenIssuer() {
...
     }
 
+    public static boolean isOrganizationValidAndActive(String organizationId) throws IdentityOAuth2Exception {
+
+        try {
+            boolean organizationExistById = OAuth2ServiceComponentHolder.getInstance().getOrganizationManager()
+                    .isOrganizationExistById(organizationId);
+            if (!organizationExistById) {
+                return false;
+            }
```
sadilchamishka, line 1436: As the authorized organization is set by the server side, it can't have invalid value. But there is a posibility the org has been deleted while processing this logic. In such case will it throw error or null. If doesn't throw an error, we can reduce above check.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
AnuradhaSK: getAuthorizedOrganization method set the authorizedOrganization to NONE if the accessing organization is null in the tokenRequestCtx
Screenshot 2024-05-28 at 17 30 30

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
AnuradhaSK: "isOrganizationExistById" returns false if the org id is invalid/ org is deleted

Is your concern is org get deleted at a point when isOrganizationExistById check is completed and it returned true, but at the point of getOrganization method is invoked the org has been deleted ?

sadilchamishka: No, Tried to remove the first method call, if the 2nd method call can be used to get org and check org exist

sadilchamishka: This is fine.

sadilchamishka (approved):

PR #2483, opened 2024-05-31
Add aud parameter in introspection response for opaque tokens

amanda-ariyaratne:
### Purpose

The aud parameter was added to the introspection response of opaque tokens by https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2457 but was later reverted due to cross-tenant introspection requests failing. This PR contains the changes of the above and corrects the application retrieval logic.

The consumer application is retrieved by the client ID and the tenant domain. Previously, the tenant domain was resolved from the thread local ie. it will resolve to be the tenant domain which the introspection call is made to. When the token is obtained for an application in a different domain, application retrieval failed. This PR retrieves the application by the client ID and the tenant domain of the access token.


piraveena (dismissed):

amanda-ariyaratne:
Changes captured in https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2485
Closing.

PR #2485, opened 2024-06-05
Add aud parameter in introspection response for opaque tokens

amanda-ariyaratne:
### Purpose

The aud parameter was added to the introspection response of opaque tokens by https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2457 but was later reverted due to cross-tenant introspection requests failing. This PR contains the changes of the above and corrects the application retrieval logic.

The consumer application is retrieved by the client ID and the tenant domain. Previously, the tenant domain was resolved from the thread local ie. it will resolve to be the tenant domain which the introspection call is made to. When the token is obtained for an application in a different domain, application retrieval failed. 

This PR retrieves the application by the client ID and the tenant domain which the application resides in. To achieve this, the new class variable `appResidentTenantId` is introduced to the Access Token model.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
@@ -28,6 +28,7 @@ import org.wso2.carbon.identity.application.common.IdentityApplicationManagement
 import org.wso2.carbon.identity.application.common.model.ServiceProvider;
 import org.wso2.carbon.identity.central.log.mgt.utils.LogConstants;
 import org.wso2.carbon.identity.central.log.mgt.utils.LoggerUtils;
+import org.wso2.carbon.identity.core.util.IdentityTenantUtil;
 import org.wso2.carbon.identity.core.util.IdentityUtil;
 import org.wso2.carbon.identity.oauth.common.OAuthConstants;
 import org.wso2.carbon.identity.oauth.common.exception.InvalidOAuthClientException;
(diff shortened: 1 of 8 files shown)


chamathns (dismissed):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/validators/TokenValidationHandler.java
```diff
@@ -862,4 +869,23 @@ private void addScopesToIntrospectionResponse(OAuth2IntrospectionResponseDTO int
...
+
+    private void addAudienceToIntrospectionResponse(OAuth2IntrospectionResponseDTO introResp,
+                                                      AccessTokenDO accessTokenDO) throws IdentityOAuth2Exception {
+
+        String tenantDomain = null;
+        try {
+            int appResidentTenantId = accessTokenDO.getAppResidentTenantId();
+            if (appResidentTenantId != MultitenantConstants.INVALID_TENANT_ID) {
+                tenantDomain = IdentityTenantUtil.getTenantDomain(appResidentTenantId);
+                OAuthAppDO oAuthAppDO = OAuth2Util.getAppInformationByClientId(accessTokenDO.getConsumerKey(),
```
indeewari, line 881: Is this nullable ?

amanda-ariyaratne: Both will be non-null values. 
Access token object is retrieved from the cache/DB.
If there is no `oAuthAppDO`, it is caught as an exception.

chamathns (approved):

PR #2510, opened 2024-07-02
Add URL validations for the token endpoint

amanda-ariyaratne:

Add URL validations to the token endpoint to prevent sending sensitive data as query parameters


components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuthConstants.java
```diff
@@ -363,6 +363,9 @@ public static class OAuth20Params {
         public static final String RESPONSE_TYPE = "response_type";
         public static final String RESPONSE_MODE = "response_mode";
         public static final String REQUEST = "request";
+        public static final String USERNAME = "username";
```
indeewari, line 366: Is it finalized that the sensitive parameters are limited to username, password, and client_id? If not, we should consider using a configuration.

amanda-ariyaratne: Added configuration with commit bbcd91132f5cea28bf3dc295ac26f3b431dc7b20

chamathns (approved):

PR #2528, opened 2024-07-25
Implement support for pre issue access token action

malithie:

Implements the facility in oauth component to implement actions for https://github.com/wso2/product-is/issues/20739

Key factors considered in the implementation as as below.
1. pre issue access token actions are allowed at tenant level and applicable for password, authorization code and client credential grant only
2. From standard access token claim set only scope, token validity period, audience are allowed to be updated
3. All OIDC claims shared in access token is allowed to be removed or updated
4. Any claims in the format of string, number, boolean or string array can be added to the access token
5. When the access token claim set is changed at token issue for password or authorization code the changed claim set is persisted in the grant cache which is backed by the session data store, to share it back with the respective service extending the token issue flow (updated context is preserved). This is valid up to the expiry of the refresh token

Further there are some minor improvements refactoring some methods and making sure to load token message context before executing the action.

Related PR: https://github.com/wso2/carbon-identity-framework/pull/5810

Once https://github.com/wso2/carbon-identity-framework/pull/5810 is merged only

### Follow up actions

1. Diagnostics logs are yet to be improved
2. Unit and integration tests are yet to be covered
3. Flows are yet to be verified for B2B cases and federated scenarios comprehensively


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
```diff
@@ -0,0 +1,719 @@
...
+/**
+ * This class is responsible for processing the response received from the action execution
+ * of the pre issue access token.
+ */
+public class PreIssueAccessTokenProcessor implements ActionExecutionResponseProcessor {
+
+    private static final Log LOG = LogFactory.getLog(PreIssueAccessTokenProcessor.class);
+    private static final String SCOPE_PATH_PREFIX = "/accessToken/scopes/";
+    private static final String CLAIMS_PATH_PREFIX = "/accessToken/claims/";
+    private static final Pattern NQCHAR_PATTERN = Pattern.compile("^[\\x21\\x23-\\x5B\\x5D-\\x7E]+$");
```
DMHP, line 56: Could not understand the reason for the name "NQCHAR_PATTERN". Is it a standard pattern?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
```diff
@@ -0,0 +1,719 @@
...
+    private static final char PATH_SEPARATOR = '/';
+
+    @Override
+    public ActionType getSupportedActionType() {
+
+        return ActionType.PRE_ISSUE_ACCESS_TOKEN;
+    }
+
+    @Override
+    public ActionExecutionStatus processSuccessResponse(ActionType actionType, Map<String, Object> eventContext,
```
DMHP, line 69: I suggest formatting the method parameters for better readability:
```suggestion
 public ActionExecutionStatus processSuccessResponse(ActionType actionType,
                                                     Map<String, Object> eventContext,
                                                     Event event,
                                                     ActionInvocationSuccessResponse actionInvocationSuccessResponseeventContext)

File: components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java [hunk 3]
```diff
@@ -0,0 +1,719 @@
...
+                                                        Event event,
+                                                        ActionInvocationSuccessResponse
+                                                                actionInvocationSuccessResponse) {
+
+        OAuthTokenReqMessageContext tokenMessageContext =
+                (OAuthTokenReqMessageContext) eventContext.get("tokenMessageContext");
```
DMHP, line 75: Add or reuse a constant for "tokenMessageContext"

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
```diff
@@ -0,0 +1,719 @@
...
+
+        //todo: need to add to diagnostic logs
+        if (LOG.isDebugEnabled()) {
+            ObjectMapper objectMapper = new ObjectMapper();
+            try {
+                String executionSummary = objectMapper.writeValueAsString(operationExecutionResultList);
+                LOG.debug(String.format("Processed response for action type: %s. Results of operations performed: %s",
+                        actionType, executionSummary));
+            } catch (JsonProcessingException e) {
+                LOG.debug("Error occurred while logging operation execution results.", e);
```
DMHP, line 123: Shall we make this as an error log?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
```diff
@@ -0,0 +1,719 @@
...
+    private OperationExecutionResult handleAddOperation(PerformableOperation operation, AccessToken requestAccessToken,
+                                                        AccessToken.Builder responseAccessToken) {
+
+        if (operation.getPath().startsWith(SCOPE_PATH_PREFIX)) {
+            return addScope(operation, responseAccessToken);
+        } else if (operation.getPath().startsWith(CLAIMS_PATH_PREFIX)) {
+            return addClaim(operation, requestAccessToken, responseAccessToken);
+        }
+
+        return new OperationExecutionResult(operation, OperationExecutionResult.Status.FAILURE,
```
DMHP, line 182: So from this, we define that the access token can not have any first-level custom attributes other than scopes or claims. Won't there be requirements to add any first-level custom attributes?

DMHP:
Reviewed Partially.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
```diff
@@ -0,0 +1,330 @@
...
+        OAuth2AccessTokenReqDTO tokenReqDTO = tokenMessageContext.getOauth2AccessTokenReqDTO();
+        AuthenticatedUser authorizedUser = tokenMessageContext.getAuthorizedUser();
+
+        PreIssueAccessTokenEvent.Builder eventBuilder = new PreIssueAccessTokenEvent.Builder();
+        eventBuilder.tenant(new Tenant(String.valueOf(IdentityTenantUtil.getTenantId(tokenReqDTO.getTenantDomain())),
+                tokenReqDTO.getTenantDomain()));
+
+        boolean isAuthorizedForUser = isAccessTokenAuthorizedForUser(tokenReqDTO.getGrantType(), tokenMessageContext);
+        if (isAuthorizedForUser) {
+            eventBuilder.organization(new Organization(authorizedUser.getUserResidentOrganization(),
```
Thumimku, line 108: Why we are passing same arg two time

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/TokenRequest.java
```diff
@@ -0,0 +1,152 @@
...
+
+            if (!headersToAvoid.contains(key.toLowerCase())) {
+                this.additionalHeaders.put(key, value);
+            }
+            return this;
+        }
+
+        public Builder addAdditionalParam(String key, String[] value) {
+
+            if (!paramsToAvoid.contains(key)) {
```
Thumimku, line 141: Let's check with out case sensitivity

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/internal/OAuthServiceComponent.java
```diff
@@ -556,4 +571,27 @@ protected void unregisterConfigurationManager(ConfigurationManager configuration
...
+    @Reference(
+            name = "action.execution.service",
+            service = ActionExecutorService.class,
+            cardinality = ReferenceCardinality.MANDATORY,
+            policy = ReferencePolicy.DYNAMIC,
+            unbind = "unregisterActionExecutorService"
+    )
+    protected void registerActionExecutionService(ActionExecutorService actionExecutorService) {
+
+        if (log.isDebugEnabled()) {
```
Thumimku, line 584: we don't need to check ```isDebugEnabled```, cause the log gonna print doesn't have any string concatenation. So checking isDebug enabled is unnecessary

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -167,6 +167,8 @@ public OAuth2AccessTokenRespDTO issue(OAuth2AccessTokenReqDTO tokenReqDTO)
         String grantType = tokenReqDTO.getGrantType();
         OAuth2AccessTokenRespDTO tokenRespDTO = null;
 
+        log.info("=== Starting access token issue flow. grantType: " + grantType);
```
Thumimku, line 170: do we need this info

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -573,6 +574,12 @@ private OAuth2AccessTokenRespDTO validateGrantAndIssueToken(OAuth2AccessTokenReq
             }
         }
 
+        log.info("=== End of access token issue flow. grantType: " + grantType + " tokenRespDTO: " + "{"
```
Thumimku, line 577: de we need this log, cause for jwt token this log will be quite big

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/OAuthTokenReqMessageContext.java
```diff
@@ -42,7 +45,7 @@ public class OAuthTokenReqMessageContext {
 
     private long refreshTokenvalidityPeriod = OAuthConstants.UNASSIGNED_VALIDITY_PERIOD;
 
-    private long accessTokenIssuedTime;
+    private Timestamp accessTokenIssuedTimestamp;
```
Thumimku, line 48: why we need this modification?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
```diff
@@ -591,8 +593,8 @@ protected JWTClaimsSet createJWTClaimSet(OAuthAuthzReqMessageContext authAuthzRe
         long curTimeInMillis = Calendar.getInstance().getTimeInMillis();
 
         AuthenticatedUser authenticatedUser = getAuthenticatedUser(authAuthzReqMessageContext, tokenReqMessageContext);
-        String sub = getSubjectClaim(consumerKey, spTenantDomain, authenticatedUser);
-        if (checkPairwiseSubEnabledForAccessTokens()) {
```
Thumimku, line 595: can we remove this implementation also?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
```diff
@@ -451,7 +457,49 @@ private OAuth2AccessTokenRespDTO generateNewAccessToken(OAuthTokenReqMessageCont
...
+            throws IdentityOAuth2Exception {
+
+        OAuthAppDO oAuthAppBean = getoAuthApp(tokenReqMessageContext.getOauth2AccessTokenReqDTO().getClientId());
+        String grantType = tokenReqMessageContext.getOauth2AccessTokenReqDTO().getGrantType();
+
+        // Allow for following grant types and for JWT access tokens only.
+        return (OAuthConstants.GrantTypes.AUTHORIZATION_CODE.equals(grantType) ||
+                OAuthConstants.GrantTypes.CLIENT_CREDENTIALS.equals(grantType) ||
+                OAuthConstants.GrantTypes.PASSWORD.equals(grantType) ||
+                OAuthConstants.GrantTypes.REFRESH_TOKEN.equals(grantType)) && "JWT".equals(oAuthAppBean.getTokenType());
```
shashimalcse, line 502: can we use a constant for JWT?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
```diff
@@ -0,0 +1,330 @@
...
+         */
+        if (tokenMessageContext.isPreIssueAccessTokenActionsExecuted()) {
+            return tokenMessageContext.getAdditionalAccessTokenClaims();
+        }
+
+        try {
+            CustomClaimsCallbackHandler claimsCallBackHandler =
+                    OAuthServerConfiguration.getInstance().getOpenIDConnectCustomClaimsCallbackHandler();
+            JWTClaimsSet claimsSet =
+                    claimsCallBackHandler.handleCustomClaims(new JWTClaimsSet.Builder(), tokenMessageContext);
```
shashimalcse, line 263: There are a couple of claims being added before calling the claims handler in the previous flow. Don't we need those claims?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/TokenRequest.java
```diff
@@ -0,0 +1,152 @@
...
+        paramsToAvoid.add("client_id");
+        paramsToAvoid.add("client_secret");
+        paramsToAvoid.add("redirect_uri");
+        paramsToAvoid.add("grant_type");
+        paramsToAvoid.add("scope");
+        // parameters from password grant
+        paramsToAvoid.add("username");
+        paramsToAvoid.add("password");
+        // parameters from refresh token grant
+        paramsToAvoid.add("refresh_token");
```
mpmadhavig, line 57: Missing sensitive params, `client_assertion`.

mpmadhavig: When these values are taken into a file config level, we need to publish that set to the developers, in order to avoid any sensitive data leakage related security issues.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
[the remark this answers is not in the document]
malithie: This is how spec defines the allowed character set for scopes [1]. This syntax means "Non-Quoted Character"

[1] https://datatracker.ietf.org/doc/html/rfc6749#appendix-A.4

malithie: Yes refer my comment at https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2528/files#r1694734929

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
malithie: This is the only place this literal is used. I don't see a requirement to define as a constant. Can incorporate though

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
malithie: This gets executed in debug scope only. So don't see a reason to indicate that as an error. 
Errors should be errors that the system really needs to be alerted on.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
malithie: Well, claims allow to add first level claims to the access token.
Something like below.

`{
  "sub": "rBG40w5KDm8NJvD_JrPh1RLR7l4a",
  "aut": "APPLICATION",
  "iss": "https://localhost:9443/oauth2/token",
  "customArray": [
    "a",
    "b",
    "c"
  ],
  "client_id": "rBG40w5KDm8NJvD_JrPh1RLR7l4a",
  "aud": [
    "example.com",
    "https://example.com/resource"
  ],
  "nbf": 1721895331,
  "customSID": "12345",
  "azp": "rBG40w5KDm8NJvD_JrPh1RLR7l4a",
  "org_id": "10084a8d-113f-4211-a0d5-efe36b082211",
  "scope": "custom-scope-1",
  "exp": 1721898931,
  "org_name": "Super",
  "iat": 1721895331,
  "jti": "8d97e5ea-adca-4f70-bd93-fa22df12af12"
}`

Note the `customArray` and `customSID`

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
malithie: This segment is yet to be improved considering b2b cases. I have forgotten to add the todo.
Will update

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
malithie: Only non standard or claims that are not specially handled are treated this way (Ideally, user's org, accessing org should be default incorporated). Will refer as special claims going forward. What this means is claims that the product defines to be incorporated to the access token without letting the developer to pick and choose from app configuration or incorporate from the extension. 
In the token processor I have a logic to filter out these special claims from the set of claims that developer has control of
https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2528/files#diff-ba5e6b8dbaba53f900aa6916029a107db0bee843cdcc3f5ff0d6e6a99d4caf15R166-R172

Hope this answers.
However, as of now, we don't have a common way of differentiating the claims in the access token that system incorporates or incorporated by the developer over some configuration or extension. Therefore, this code block handles such specially and that code segment is duplicated in JWTTokenIssuer as well

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/TokenRequest.java
malithie: Well request parameters are case sensitive (Refer thread below that indicates such). So, IMO we should not treat is as case insensitive.

https://stackoverflow.com/questions/8969170/httpservletrequest-parameter-lowercase

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/internal/OAuthServiceComponent.java
malithie: ack. will remove

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/OAuthTokenReqMessageContext.java
malithie: This was done, so that we update the message context with the issued time and later we read from message context and set back to the token.
However it doesn't require to change this field to a Timestamp field. Will update with existing fields

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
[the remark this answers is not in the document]
malithie: Well, the formatter doesn't format so. IMO, we don't need to take additional overhead in such cases

malithie: My same comment above applies

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/internal/OAuthServiceComponent.java
malithie: Fixed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/OAuthTokenReqMessageContext.java
malithie: Reverted back still setting the issued time from message context

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/JWTTokenIssuer.java
malithie: removed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/TokenRequest.java
malithie: Fixed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
shashimalcse: ack

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenProcessor.java
```diff
@@ -0,0 +1,719 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * This class is responsible for processing the response received from the action execution
+ * of the pre issue access token.
+ */
+public class PreIssueAccessTokenProcessor implements ActionExecutionResponseProcessor {
```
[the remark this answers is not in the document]
malithie: Fixed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
malithie: removed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AbstractAuthorizationGrantHandler.java
malithie: fixed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/OAuth2Util.java
```diff
@@ -5527,4 +5529,9 @@ private static void addOrganizationUserDetails(AuthenticatedUser authenticatedUs
         // Set authorized user tenant domain to the tenant domain of the application.
         authenticatedUser.setTenantDomain(appTenantDomain);
     }
+
+    public static boolean isPairwiseSubEnabledForAccessTokens() {
```
DMHP, line 5533: Add a method comment

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
```diff
@@ -667,6 +678,72 @@ private static void addUserAttributesToCache(AccessTokenDO accessTokenBean,
...
+    private boolean checkExecutePreIssueAccessTokensActions(RefreshTokenValidationDataDO refreshTokenValidationDataDO,
+                                                            OAuthTokenReqMessageContext tokenReqMessageContext)
+            throws IdentityOAuth2Exception {
+
+        OAuthAppDO oAuthAppBean = getOAuthApp(tokenReqMessageContext.getOauth2AccessTokenReqDTO().getClientId());
+        String grantType = refreshTokenValidationDataDO.getGrantType();
+
+        // Allow if refresh token is issued for token requests from following grant types and,
+        // for JWT access tokens only.
+        return (OAuthConstants.GrantTypes.AUTHORIZATION_CODE.equals(grantType) ||
```
DMHP, line 722: Any reason to stick to this particular grant types only?

malithie: This delivery focuses on these grant types only.

So a specific check was added to make sure this gets engaged for tested grant types and nothing breaks if it's a different grant type.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
```diff
@@ -667,6 +678,72 @@ private static void addUserAttributesToCache(AccessTokenDO accessTokenBean,
...
+
+        AuthorizationGrantCacheKey grantCacheKey = new AuthorizationGrantCacheKey(refreshTokenData.getTokenId());
+        AuthorizationGrantCacheEntry grantCacheEntry = AuthorizationGrantCache.getInstance()
+                .getValueFromCacheByTokenId(grantCacheKey, refreshTokenData.getTokenId());
+
+        if (grantCacheEntry != null && grantCacheEntry.isPreIssueAccessTokenActionsExecuted()) {
+            tokenRequestContext.setPreIssueAccessTokenActionsExecuted(true);
+            tokenRequestContext.setAdditionalAccessTokenClaims(grantCacheEntry.getCustomClaims());
+            tokenRequestContext.setAudiences(grantCacheEntry.getAudiences());
+            if (log.isDebugEnabled()) {
```
DMHP, line 738: if (log.isDebugEnabled()) redundant

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/OperationExecutionResult.java
```diff
@@ -0,0 +1,44 @@
+package org.wso2.carbon.identity.oauth.action.model;
```
DMHP, line 1: license header is missing

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/AccessToken.java
```diff
@@ -0,0 +1,256 @@
...
+            } else if (value instanceof Map) {
+                Map<?, ?> originalMap = (Map<?, ?>) value;
+                Map<Object, Object> copiedMap = new HashMap<>();
+                for (Map.Entry<?, ?> entry : originalMap.entrySet()) {
+                    copiedMap.put(entry.getKey(), deepCopyValue(entry.getValue())); // Recursive for nested maps
+                }
+                return copiedMap;
+            } else if (value.getClass().isArray()) {
+                return Arrays.copyOf((Object[]) value, ((Object[]) value).length);
+            } else {
```
DMHP, line 177: This else is redundant. We can just add  "// For immutable types or types not requiring deep copy
                return value; " without specifying the else block

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
```diff
@@ -0,0 +1,352 @@
...
+        return eventBuilder.build();
+    }
+
+    private void setUserForEventBuilder(PreIssueAccessTokenEvent.Builder eventBuilder, AuthenticatedUser user,
+                                        String clientID, String grantType) {
+
+        try {
+            eventBuilder.user(new User(user.getUserId()));
+        } catch (UserIdNotFoundException e) {
+            if (LOG.isDebugEnabled()) {
```
DMHP, line 128: Are we not breaking the flow even if the 'user id' is not found when the access token us authorized for user?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/ClaimPathInfo.java
```diff
@@ -0,0 +1,45 @@
...
+ * under the License.
+ */
+
+package org.wso2.carbon.identity.oauth.action.model;
+
+/**
+ * This class is used to identify the claim and the index of the value of the claim to be updated in the
+ * {@link AccessToken}. The claim value of the {@link AccessToken} is an array in this case.
+ */
+public class ClaimPathInfo {
```
Thumimku, line 25: I think ```Claim``` may be applicable class name, given that attributes explain the claim in the AT

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
malithie: ack

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
malithie: No, IMO user id is not found it should fall back to a username or something. Need to verify this behaviour more and improve.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/AccessToken.java
malithie: Since there is an else if, this else is needed. 
It's not redundant and the IDE doesn't report.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/ClaimPathInfo.java
malithie: This is not a claim. Basically this rather identifies the claim name in the path redefined at a particular operation in response.

e.g:

{
      "op": "replace",
      "path": "/accessToken/claims/aud/-",
      "value": "example.com"
    }

Here when above operation is processed, this class is used to identify the claim name in the path which is "aud" and the index that's defined with respect to that claim which is "-" that denotes the last index

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
malithie: Added a todo so we can improve later

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/OperationExecutionResult.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.wso2.carbon.identity.action.execution.model.PerformableOperation;
+
+/**
+ * This class represents the result of the execution of an operation.
+ * It contains the operation that was executed, the status of the execution and a message.
+ * This is used to summarize the operations performed based on action response.
+ */
+public class OperationExecutionResult {
```
Thumimku, line 28: Since this is generic class please check whether we can move this class to the event core and reuse for future implementations

malithie: Agree. Will move as we progress for one or two more extensions. It's fine to break these contracts as we are still under development

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/TokenRequest.java
```diff
@@ -0,0 +1,155 @@
...
+        // parameters used for client authentication for token endpoint
+        paramsToAvoid.add("client_id");
+        paramsToAvoid.add("client_secret");
+        paramsToAvoid.add("client_assertion_type");
+        paramsToAvoid.add("client_assertion");
```
Thumimku, line 60: Let's add cnf claim also for Dpop

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
```diff
@@ -0,0 +1,354 @@
...
+                // todo: fall back to a different identifier like username.
```
DMHP, line 129: Resolving organization flow also should be improved.

DMHP:
Due to the priority, I am approving the following PR with the following,
- [ ] As discussed with @malithie she has immediate plans to onboard the unit tests to cover the whole flow
- [ ] This feature will be gradually improved to all the grant types including custom grant types.
- [ ] Some TODOs will be completed along with the feature.

DMHP (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
```diff
@@ -0,0 +1,354 @@
...
+
+        return ActionType.PRE_ISSUE_ACCESS_TOKEN;
+    }
+
+    @Override
+    public ActionExecutionRequest buildActionExecutionRequest(Map<String, Object> eventContext)
+            throws ActionExecutionRequestBuilderException {
+
+        OAuthTokenReqMessageContext tokenMessageContext =
+                (OAuthTokenReqMessageContext) eventContext.get("tokenMessageContext");
```
Thumimku, line 86: We are using "tokenMessageContext" in other place also, can we make it as constant.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/TokenRequest.java
malithie: Does that claim goes as a request param ? It rather looks like we need to avoid the DPoP header "DPoP"

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
malithie: Well we don't have a common constant class for actions in oauth component scope. Just for this I'm bit reluctant to add a class and add this or to move which is specific to actions to a common constant class of OAuth

malithie:
> Due to the priority, I am approving the following PR with the following,
> 
> * [ ]  As discussed with @malithie she has immediate plans to onboard the unit tests to cover the whole flow
> * [ ]  This feature will be gradually improved to the token exchange grant based on 80% cases.
> * [ ]  The TODOs will be completed along with the feature.

Acknowleged.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/PreIssueAccessTokenRequestBuilder.java
malithie: Well that won't have an option. If there's an error in resolving org request will not contain the org, which is fine.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/action/model/TokenRequest.java
Thumimku: Yes, we need to avoid "DPoP" header. but here we are talking about params and headers so ignore my previous claim. CNF claim will be in AT.

PR #2555, opened 2024-08-22
Fix refresh token expiry time issue

amanda-ariyaratne:

Fix converting the callback validity period to millisecond, since its already set in milliseconds.
This change is an immediate fix to avoid increasing the token expiry time exponentially. The complete solution needs to be revisited.


amanda-ariyaratne:
Closing because duplicate of https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2556

PR #2568, opened 2024-09-09
Fix internal server error when JSON string used as state param

SujanSanjula96:
**Purpose**
Fix internal server error when JSON string used as state param in authorize request. This is fixed for both query and fragment response modes.

Issue - https://github.com/wso2/product-is/issues/20919

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/responsemode/provider/impl/QueryResponseModeProvider.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/responsemode/provider/impl/QueryResponseModeProvider.java
@@ -19,6 +19,8 @@
 package org.wso2.carbon.identity.oauth2.responsemode.provider.impl;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


sadilchamishka (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/responsemode/provider/impl/QueryResponseModeProvider.java
```diff
@@ -154,4 +160,21 @@ public AuthResponseType getAuthResponseType() {
...
+    }
+
+    private String encodeValue(String value) {
+
+        try {
+            return URLEncoder.encode(value, StandardCharsets.UTF_8.toString());
+        } catch (UnsupportedEncodingException e) {
+            // This exception will not be thrown as UTF-8 is always supported.
+            log.error("Error occurred while encoding the value: " + value, e);
+            return null;
```
sadilchamishka, line 177: Returning null might case NPE when concat with strings.
Let's fix it in a later PR.

SujanSanjula96: This exception will not be thrown since UTF-8 is always supported.

PR #2109, opened 2023-06-28
Re-introduce session state param condition

Yoshani:
### Proposed changes in this pull request

With the changes introduced by PR [1], in the manageOIDCSessionState() method the [if condition](https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/3d20187ca2db291822e257511c2e88e408088086/components/org.wso2.carbon.identity.oauth.endpoint/src/main/java/org/wso2/carbon/identity/oauth/endpoint/authz/OAuth2AuthzEndpoint.java#L3384) that was present previously has been missed. In the device code grant flow, no redirect url is passed. Therefore this resulted in an NPE.
This PR reintroduces this if condition to resolve the issue.

### Related Issues
- Issue https://github.com/wso2/product-is/issues/16139

### Related PRs

- [1] PR https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2037

PR #2639, opened 2024-12-03
Exclude OIDC Session Management for Secondary Device in Device Grant Flow

SujanSanjula96:
From this PR, OIDC Session Management is excluded for Secondary Device in Device Grant Flow. With that, NPE when getting session state param due to `null` callback URL is fixed. 

Issue - https://github.com/wso2/product-is/issues/16867


SujanSanjula96:
Already fixed with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2109

PR #2664, opened 2025-01-07
Inherit parent user domain in shared token revoke flow

HasiniSama:
### Purpose
When revoking an access token in a shared user flow, the authorized user's **User Store Domain** is set to the shared user's domain. Therefore, when retrieving the clientIDs from [here](https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/TokenManagementDAOImpl.java#L877), it will only retrieve the clientIDs associated with the shared user's User Store Domain. 

However, when creating access tokens for shared users the User Store Domain of the parent user is used [here](https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java#L164) and [here](https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java#L219). Hence these tokens are not getting revoked. 

### Approach
Hence the access token revoke logic for shared user flow has to be improved to handle the user's domain correctly which will eventually detect all the clientIDs issued.

This is an alternative approach to the fix https://github.com/wso2-extensions/identity-oauth2-grant-organization-switch/pull/38. In which we amend the user store domain when the token is issued. Which would lead to user store mismatch between the token and the actual user. Here, we change the logic in the token revoke method.

### Tested Flows (Unit test added for the relevant flows)

| Scenario                        | User Domain  | Authz User           | Tenant | Authorized Organization | Grant Type            | IDP_ID | Unit Test Added |
|---------------------------------|--------------|----------------------|--------|--------------------------|-----------------------|--------|-----------------|
| Initial login with no switch    | Primary      | Username             | -1234  | NONE                     | authorization_code    | 1      | ❌               |
| Switch to OrgID_01              | Primary      | Username             | -1234  | OrgID_01                 | organization_switch   | 1      | ✅               |
| Switch to OrgID_02              | Primary      | Username             | -1234  | OrgID_02                 | organization_switch   | 1      | ✅               |
| Initial login with no switch    | Secondary    | Username             | -1234  | NONE                     | authorization_code    | 1      | ❌               |
| Switch to OrgID_01              | Secondary    | Username             | -1234  | OrgID_01                 | organization_switch   | 1      | ✅               |
| Switch to OrgID_02              | Secondary    | Username             | -1234  | OrgID_02                 | organization_switch   | 1      | ✅               |
| SSO login to OrgID_01           | Primary*     | Username             | 33     | NONE                     | authorization_code    | 33     | ❌               |
| SSO login to OrgID_01           | Federated    | User ID              | 33     | OrgID_01                 | authorization_code    | 3      | ❌               |
| Switch to OrgID_02 after SSO    | Federated    | User ID              | 33     | OrgID_02                 | organization_switch   | 3      | ✅               |
| Federated login without JIT     | Federated    | Federated Username   | -1234  | NONE                     | authorization_code    | 38     | ❌               |
| JIT provisioned Federated login | Secondary    | JIT given Username   | -1234  | NONE                     | authorization_code    | 1      | ❌               |

*The access token issued by the shared application to the parent application. This doesn't get deleted from the database when the user logs out.


SujanSanjula96 (dismissed):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
```diff
@@ -1334,4 +1339,41 @@ private static String readServerConfigurationPvtKeyJWTReuse() {
...
+     * @return Optional containing the parent user's user store domain, or empty if not found.
+     * @throws OrganizationManagementException If an error occurs retrieving user association.
+     * @throws UserStoreException              If an error occurs retrieving the user store domain.
+     */
+    private static Optional<String> getUserStoreDomainOfParentUser(String userId, String accessingOrgId,
+                                                                   String tenantDomain)
+            throws OrganizationManagementException, UserStoreException {
+
+        String parentUserId = OAuthComponentServiceHolder.getInstance().getOrganizationUserSharingService()
+                .getUserAssociation(userId, accessingOrgId)
```
AnuradhaSK, line 1358: Any chance of returning null from getUserAssociation(...) ? If yes, do a null check to avoid NPE

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
```diff
@@ -747,6 +748,10 @@ private static AuthenticatedUser buildAuthenticatedUser(UserStoreManager userSto
                 authenticatedUser.setUserResidentOrganization(managedOrg);
                 authenticatedUser.setAccessingOrganization(accessingOrg);
 
+                Optional<String> parentUserStoreDomain = getUserStoreDomainOfParentUser(
+                        userId, accessingOrg, tenantDomain);
+                parentUserStoreDomain.ifPresent(authenticatedUser::setUserStoreDomain);
```
sadilchamishka, line 753: This part may need to move to below else block.
Shall we try the flow for sub-org and another level of sub-org. No need to add a user store.

AnuradhaSK (dismissed):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
HasiniSama: This is now updated to null check `getUserAssociation()`.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
HasiniSama: This is now refactored only to be triggered for local login flow.

lashinijay (approved):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
```diff
@@ -1327,4 +1333,43 @@ private static String readServerConfigurationPvtKeyJWTReuse() {
...
+        String parentUserId = userAssociation.getAssociatedUserId();
+
+        try {
+            int tenantId = IdentityTenantUtil.getTenantId(tenantDomain);
+            UserRealm userRealm = OAuthComponentServiceHolder.getInstance()
+                    .getRealmService()
+                    .getTenantUserRealm(tenantId);
+            UserStoreManager userStoreManager = (AbstractUserStoreManager) userRealm.getUserStoreManager();
+
+            return Optional.ofNullable(((AbstractUserStoreManager) userStoreManager)
```
ashanthamara, line 1367: Is it expected to be the parent userstore domain to be null ?
Shouldn't be handled as an error ?

HasiniSama: Refactored via: https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2787

PR #2666, opened 2025-01-08
Fix DCR name update

JeethJJ:
DCR sp object updated twice will remove the name update. Hence this PR refines the logic to update the object from one point.


Thumimku (approved):
Please run the PR builder

PR #2669, opened 2025-01-09
Fix DCR PUT request not updating application role audience

Thumimku:
This PR fixes an issue in the DCR PUT implementation where the application role audience value was not persisted to the database.

### Changes

- Added logic to store the application role audience during DCR updates.
- Earlier if the client name is not is the request the SP is not updated, thats not correct, hence moving the update SP section outside the if block.
- Improved unit tests to assert the application role audience update.

### Issue

https://github.com/wso2/product-is/issues/21627


JeethJJ (approved):

Thumimku:
Associated failure not related to this PR

PR #2674, opened 2025-01-16
Improve user info response to return roles without internal domain appended

sadilchamishka:

Add the below config to enable this improvement.
```
[oauth.oidc.user_info]
remove_internal_prefix_from_roles=true
```
### Related Issues
- https://github.com/wso2/product-is/issues/22088


shashimalcse (dismissed):

lashinijay (approved):

sadilchamishka:
The added test cases cover the flow. But the test classes written in the oauth endpoint module. Even the required case is validated, the code coverage is not counted as the code is covered from the a different mdoule. 
Merging the PR as the logic is covered via the added test case.

PR #2683, opened 2025-01-22
Remove commonAuthId cookie if authentication flow failed

hwupathum:
### Summray
This pull request includes an important change to the `OAuth2AuthzEndpoint` class in the `components/org.wso2.carbon.identity.oauth.endpoint` package. The change ensures that the `commonAuthId` cookie is removed if the authentication flow fails.

### Related issue
 - https://github.com/wso2/product-is/issues/22194


hwupathum:
Decided not send this fix since this is an edge case

PR #2506, opened 2024-06-27
Add support for additional attributes in the DCR endpoint

amjadhifthikar:
### Proposed changes in this pull request

- This PR introduces the support for additional attributes in the DCR endpoint. Additional attributes can be sent in the registration request, updated in the DCR update request, and retrieved in the get request. Additional attributes can be filtered, validated, and stored as required. More details on the extension is available in the interface comments.

PR #2689, opened 2025-01-24
Fix issue in DCR Additional attribute filter not working as OSGI

Ashi1993:

This PR fix issue in DCR Additional attribute filter added via [1] not working as OSGI

Related issue : https://github.com/wso2/product-is/issues/22404

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2506

-

can merge this pull request.]

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.dcr/src/main/java/org/wso2/carbon/identity/oauth/dcr/service/DCRMService.java
+++ b/components/org.wso2.carbon.identity.oauth.dcr/src/main/java/org/wso2/carbon/identity/oauth/dcr/service/DCRMService.java
@@ -27,7 +27,6 @@ import org.apache.commons.logging.LogFactory;
 import org.wso2.carbon.context.CarbonContext;
 import org.wso2.carbon.context.PrivilegedCarbonContext;
 import org.wso2.carbon.identity.application.common.IdentityApplicationManagementException;
-import org.wso2.carbon.identity.application.common.IdentityApplicationManagementValidationException;
 import org.wso2.carbon.identity.application.common.model.AssociatedRolesConfig;
 import org.wso2.carbon.identity.application.common.model.InboundAuthenticationConfig;
 import org.wso2.carbon.identity.application.common.model.InboundAuthenticationRequestConfig;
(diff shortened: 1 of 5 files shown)


Thumimku (approved):

components/org.wso2.carbon.identity.oauth.dcr/src/main/java/org/wso2/carbon/identity/oauth/dcr/service/DCRMService.java
```diff
@@ -120,18 +117,8 @@ public Application getApplication(String clientId) throws DCRMException {
         Application application = buildResponse(consumerAppDTO, tenantDomain);
         application.setExtAllowedAudience(serviceProvider.getAssociatedRolesConfig().getAllowedAudience());
 
-        String attributeFilterName = IdentityUtil.getProperty(OAuthConstants.ADDITIONAL_ATTRIBUTE_FILTER);
```
nilasini, line 123: why are we removing the identity.xml file reading? Isn't there any case where customers have the customization which is not an OSGI service?

Ashi1993: Even though the config is there, it is not working in the latest update levels as well. Hence, it won't be working for customers as well.

Ashi1993:
Even though the config is there, it is not working in the latest update levels as well. Hence it won't be working for customers as well. SO removing the config won't be an issue.

components/org.wso2.carbon.identity.oauth.dcr/src/main/java/org/wso2/carbon/identity/oauth/dcr/service/DCRMService.java
nilasini: Was this working in the default 7.0.0 pack? Don't we need to do the fix in a way that should work with non-osgi and osgi?

Ashi1993: This was added as a patch for IS 7.0. And was not working when the extension was from another bundle. Hence I dont think there is an issue

Thumimku (approved):

components/org.wso2.carbon.identity.oauth.dcr/src/main/java/org/wso2/carbon/identity/oauth/dcr/service/DCRMService.java
nilasini: Ok does that meant this has never worked before? Even when it was added initially?

Ashi1993: Yes thats correct. But if the extension is in the same module it works.

PR #2700, opened 2025-02-08
Correct errorMessage and errorCode sent in pw grant accountlock resp

mpmadhavig:

Previously the error_description & error combination for the account lock response for oauth apps has been sent as below.
```
{
    "error_description": "17003:AdminInitiated Error when handling event : POST_AUTHENTICATION",
    "error": "invalid_grant"
}
```

The root cause for this issue are as below.

error_description
- The error_description is being set using the errorMessage field of an execption.
- When creating the error message, instead of getting the AccountLockException errorMessage, it has used more generic UserStoreException errorMessage.

error
- error is set using the errorCode field of an execption.
- As the error code is not set during the error throwing stage [1], the more generic error `invalid_grant` has been sent [2].

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/8001c6786a92cac48fa28bbfacf351598bcacce2/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/PasswordGrantHandler.java#L450
[2] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/8001c6786a92cac48fa28bbfacf351598bcacce2/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java#L398

After fixing the above issues, the error response will look like below.

```
{
    "error_description": "Account is locked by admin for user: a*****r in user store: PRIMARY in tenant: carbon.super. Cannot login until the account is unlocked.",
    "error": "17003:AdminInitiated"
}
```

Resolves https://github.com/wso2/product-is/issues/22585


ShanChathusanda93 (dismissed):

mpmadhavig:
During a discussion with @DMHP it was identified that the error field is set as the `invalid_grant` is intentional. As per the OAuth 2.0 spec [1], the error field must include one the defined types in the spec. 

Among the defined errors `invalid_grant` is the most suitable error as it communicates that the credentials cannot be used to authenticate due to a condition that invalidates their usability.

```
invalid_grant
               The provided authorization grant (e.g., authorization
               code, resource owner credentials) or refresh token is
               invalid, expired, revoked, does not match the redirection
               URI used in the authorization request, or was issued to
               another client.
```

Therefore removing the error field changes.

[1] https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

mpmadhavig:
Please refer https://github.com/mpmadhavig/product-is/actions/runs/13213686129

mpmadhavig:
> Please refer https://github.com/mpmadhavig/product-is/actions/runs/13213686129

The build has passed.

DMHP (approved):

PR #2713, opened 2025-02-15
Upgrade nimbusds version to 9.37.3 version

sandushi:
- $subject
- Change the code because now the getJSONObjectClaim returns Map<String, Object> instead of JSONobject with the new version
- toJSONObject method also now returns Map<String, Object> instead of JSONobject

### Referenced PR
https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2391

### Related issue
https://github.com/wso2/product-is/issues/23054


nilasini (dismissed):

sandushi:
The PR is still open because this is a major version upgrade and need more testing effort, will merge this PR after the testing effort.

piraveena:
Will do this in a separate effort. Hence closing this

PR #2728, opened 2025-02-21
Improve active and inactive access token retrieval queries for MSSQL when con-app-key violation occurred

sadilchamishka:

When con-app-key constrain violation occurred, following two queries get executed. 
As per the analysis done for the MSSQL database, we can notice that it takes considerable amount of time to invoke the both queries (roughly 300ms per each). 

Screenshot 2025-02-21 at 08 15 52

The reason for the latency is due to the absence of NO LOCK syntax in the queries.

We have been using the same pattern when checking the existence of a token[2],[3]  but we have missed in this place. 

[1] - https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java#L2598~L2605
[2] - https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/9db893bf03014e0922b504c10b7e2bcf2d23726c/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/SQLQueries.java#L205
[3] - https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/9db893bf03014e0922b504c10b7e2bcf2d23726c/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/SQLQueries.java#L213


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/SQLQueries.java
```diff
@@ -1172,6 +1172,7 @@ public class SQLQueries {
     public static final String RETRIEVE_LATEST_ACTIVE_ACCESS_TOKEN_BY_CLIENT_ID_USER_SCOPE_IDP_NAME_MSSQL =
             "SELECT TOP 1 ACCESS_TOKEN, REFRESH_TOKEN, TIME_CREATED, REFRESH_TOKEN_TIME_CREATED, VALIDITY_PERIOD, " +
             "REFRESH_TOKEN_VALIDITY_PERIOD, USER_TYPE, TOKEN_ID, SUBJECT_IDENTIFIER FROM IDN_OAUTH2_ACCESS_TOKEN " +
+            "WITH (NOLOCK) " +
```
indeewari, line 1175: When con_app_key is supposed to prevent persisting the same token by multiple threads by synchronizing,  how could removing this search lock not lead to data inconsistency ?

sadilchamishka: There is no point of having a read lock for the access-token table. Even we retrieve a data with a read lock, when we take the decision and take the action, the state might have been changed and again fail with con-app-key constraint violation.

The con-app-key violation reaching max attempt (default value 5) will be anyway reach having this improvement or not, but this improvement decrease the time to of reaching to the max failure attempts.

This limitation is only for the SQL server setups, as other databases works in the read with no lock status.

Thumimku: I agree that using a `NOLOCK` query will increase concurrent reads, but since the **AT table** is crucial, I am thinking whether allowing dirty reads is a tradeoff we can consider for performance improvement?

sadilchamishka: The dirty read happens for other database types also. This approach has been taken by SQL server as its default behaviour. If @Thumimku your argument is correct we have to fix the queries for other database types.

There is no any gain we read a data with a lock. Here we interested in check the existence. It just an idea that we are ready to do an insert. It is not an guarantee that we can do the insert. Hence a dirty read is fine for such a usecase.

PR #2732, opened 2025-02-22
Improve the claim retrieval for JWT tokens

madurangasiriwardena:
Improve the fix done for https://github.com/wso2/product-is/issues/20517 with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2503 to avoid unnecessary DB calls.

PR #2733, opened 2025-02-24
Handle token null scenario when retrieving the token hash from the issuer

madurangasiriwardena:
Related to the fix in https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2732


shashimalcse (approved):

Thumimku:
The flow was working for code grant but not working for password grant. I tested the flow with the fix and its working .

PR #2737, opened 2025-02-26
Set requestBinding for refresh token grant.

Yasasr1:
### Purpose

Sets the requestBinding type to the tokens issued from the refresh token grant when the renew_token_without_revoking_existing config is enabled and refresh token grant is set as an allowed grant type

### Related issue
- https://github.com/wso2/product-is/issues/23146


piraveena (approved):

piraveena:
need to add tests :)

PR #2848, opened 2025-08-11
Enhance cross-sub-organization access token introspection handling

KaveeshaPiumini:
### Purpose
The purpose of this fix is to ensure consistent behavior when introspecting opaque access tokens in a multi-tenant environment.

### Implementation
The fix refines the token introspection logic by adding an explicit check for sub-organization tokens during the validation process. 

Previously, the logic included a check to determine whether the application was created within a sub-organization. If so, it would set the tenant domain to the application's tenant domain, effectively bypassing the introspecting tenant. This behavior has been removed, as it provided no meaningful benefit for the introspection process.

In sub-organization scenarios, the accessTokenDO already contains an accessing organization, which must be handled explicitly in the validation flow.

Following configuration ensures that when the access token belongs to a sub-organization, introspection is only allowed if cross-tenant introspection for token issued for sub orgs is enabled in the deployment.toml.
```
[oauth.introspect]  
allow_cross_sub_org = true  
```

### Related PRs
- https://github.com/wso2/carbon-identity-framework/pull/7092

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
@@ -182,7 +182,7 @@ public class OAuthServerConfiguration {
     private boolean accessTokenPartitioningEnabled = false;
     private boolean redirectToRequestedRedirectUriEnabled = true;
     private boolean allowCrossTenantIntrospection = true;
-    private boolean allowCrossSubOrgIntrospection = false;
+    private boolean allowCrossTenantIntrospectionForSubOrgTokens = false;
     private boolean useClientIdAsSubClaimForAppTokens = true;
     private boolean removeUsernameFromIntrospectionResponseForAppTokens = true;
     private boolean useLegacyScopesAsAliasForNewScopes = false;
@@ -559,7 +559,7 @@ public class OAuthServerConfiguration {
         parseAllowCrossTenantIntrospection(oauthElem);
 
         // Read config for cross sub org allow.
-        parseAllowCrossSubOrgIntrospection(oauthElem);
+        parseAllowCrossTenantIntrospectionForSubOrgTokens(oauthElem);
 
         // Read config for using client id as sub claim for application tokens.
         parseUseClientIdAsSubClaimForAppTokens(oauthElem);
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -3939,6 +3943,21 @@ private void parseAllowCrossTenantIntrospection(OMElement oauthConfigElem) {
         }
     }
 
+    /**
+     * Parses the AllowCrossTenantTokenIntrospection configuration that used to allow or block token introspection
```
AnuradhaSK, line 3947: change AllowCrossTenantTokenIntrospection part

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
```diff
@@ -182,6 +182,7 @@ public class OAuthServerConfiguration {
     private boolean accessTokenPartitioningEnabled = false;
     private boolean redirectToRequestedRedirectUriEnabled = true;
     private boolean allowCrossTenantIntrospection = true;
+    private boolean allowCrossSubOrgIntrospection = false;
```
AnuradhaSK, line 185: It would be better to rename the config and variables to allowCrossTenantIntrospectionForSubOrgTokens

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
KaveeshaPiumini: Fixed with https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2859

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
KaveeshaPiumini: Fixed with d196ce7d6b6fa95d14ec61d2b09709ced52637f9

PR #2860, opened 2025-08-15
Add filter for application-specific role filtering in JWT access token

thiva-k:
**Purpose**

Add a filter to include only application-specific roles when building JWT access token with roles claim retrieved from cache. This doesn't modify the existing caching implementation.

**Related Issue**

- https://github.com/wso2/product-is/issues/24215


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/JWTAccessTokenOIDCClaimsHandler.java
```diff
@@ -1016,4 +1026,202 @@ private boolean isPreserverClaimUrisInAssertion(OAuthTokenReqMessageContext requ
...
+                        if (log.isDebugEnabled()) {
+                            log.debug("No application-associated roles found for user " + 
+                                    authenticatedUser.getLoggableMaskedUserId() + " in application " + clientId);
+                        }
+                    }
+                }
+            }
+            
+            // Also handle APP_ROLES_CLAIM if present
+            if (claims.containsKey(APP_ROLES)) {
```
SujanSanjula96, line 1088: I think there is no possibility for this scenario since we only add app_roles claim with filtered list

PR #2889, opened 2025-09-01
Extend Authorization Grant Cache Validity to Refresh Token Expiry

hwupathum:
**Related Issue**
- https://github.com/wso2/product-is/issues/25136

**Background**

- When federated users authenticate with JIT provisioning disabled, their federated claims are stored only in the session store and Authorization Grant cache.
- Once the cache entry expires (based on access token validity) and the session store entry is cleaned up, there is no longer any data to retrieve these claims.
- As a result, refresh token requests issued after the cache expiry will return tokens without federated claims.

**Root Cause**
- Default cache validity is tied to access token expiry.
- Federated claims are lost if the cache is cleaned before the refresh token expiry.
- During a refresh token grant, claims are loaded from the cache; if missing, they are not re-fetched from the federated IdP.

**Solution**
- This fix introduces a configurable option to extend the Authorization Grant cache validity to match the refresh token expiry.
- Default behavior (config disabled): Cache validity = Access token validity.
- With config enabled: Cache validity = Refresh token validity.

**Implementation details:**
- Updated addToCacheByToken() to set cache validity period based on the new config.
- If AccessTokenDO is not available, modified setExpiryInMillis in responseDTO (used only for cache validity, not the response expiresIn value) to reflect the updated expiry period.

**Reproduction Steps (Before Fix)**
- Configure a second WSO2 IS instance as a federated IdP.
- Request some federated claims from that IdP.
- Authenticate and verify the ID token contains the federated claims.
- In the embedded H2 DB, delete the AuthorizationGrantCache entries for the code and token.
- Restart IS.
- Perform a refresh token grant.
- Observe: New tokens do not contain federated claims.

**Verification (After Fix)**
- Enable the config in deployment.toml.
- Repeat the reproduction steps.
- Federated claims remain available in tokens until the refresh token expires.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/authz/handlers/TokenResponseTypeHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/authz/handlers/TokenResponseTypeHandler.java
@@ -579,8 +579,9 @@ public class TokenResponseTypeHandler extends AbstractResponseTypeHandler {
                 }
                 authorizationGrantCacheEntry.setSubjectClaim(sub);
             }
-            authorizationGrantCacheEntry
-                    .setValidityPeriod(TimeUnit.MILLISECONDS.toNanos(accessTokenDO.getValidityPeriodInMillis()));
+            // Setting the validity period of the cache entry to be same as the validity period of the refresh token.
+            authorizationGrantCacheEntry.setValidityPeriod(
+                    TimeUnit.MILLISECONDS.toNanos(accessTokenDO.getRefreshTokenValidityPeriodInMillis()));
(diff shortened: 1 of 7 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/authz/handlers/TokenResponseTypeHandler.java
```diff
@@ -579,8 +579,9 @@ private void addUserAttributesToCache(String accessToken,
                 }
                 authorizationGrantCacheEntry.setSubjectClaim(sub);
             }
-            authorizationGrantCacheEntry
```
VivekVinushanth, line 582: Shouldn't this be configuration based?

VivekVinushanth (dismissed):

NipuniBhagya (approved):

PR #2901, opened 2025-09-09
Fix issue with stale user claims being returned on token refresh after access token expiry

AfraHussaindeen:

- $Subject
- Resolves https://github.com/wso2/product-is/issues/25570

## Developer Checklist (Mandatory)

- [x] Complete the **Developer Checklist** in the related `product-is` issue to track any behavioral change or migration impact.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
```diff
@@ -1837,31 +1857,36 @@ public Set<AccessTokenDO> getAccessTokensByUserForOpenidScope(AuthenticatedUser
...
-            IdentityOAuth2Exception {
+                ResultSet rs, boolean includeExpiredAccessTokensWithActiveRefreshToken)
+            throws SQLException, IdentityOAuth2Exception {
 
         Map<String, AccessTokenDO> tokenMap = new HashMap<>();
         while (rs.next()) {
             Timestamp timeCreated = rs.getTimestamp(4, Calendar.getInstance(TimeZone.getTimeZone(UTC)));
             long issuedTimeInMillis = timeCreated.getTime();
             long validityPeriodInMillis = rs.getLong(5);
+            Timestamp refreshTokenTimeCreated = rs.getTimestamp(6, Calendar.getInstance(TimeZone.getTimeZone(UTC)));
```
piraveena, line 1868: do we need to get this information if `includeExpiredAccessTokensWithActiveRefreshToken` is false?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
```diff
@@ -1821,7 +1840,8 @@ public Set<AccessTokenDO> getAccessTokensByUserForOpenidScope(AuthenticatedUser
             ps.setString(5, OAuthConstants.Scope.OPENID);
             rs = ps.executeQuery();
 
-            Map<String, AccessTokenDO> tokenMap = getAccessTokenDOMapFromResultSet(authenticatedUser, rs);
+            Map<String, AccessTokenDO> tokenMap = getAccessTokenDOMapFromResultSet(authenticatedUser, rs,
+                    includeExpiredAccessTokensWithActiveRefreshToken);
```
piraveena, line 1844: How this token became expired? Normally tokens become expired when we issue new token. As per the scenario you mentioned in the issue, it looks like token refresh happens after user profile update.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
AfraHussaindeen: if `includeExpiredAccessTokensWithActiveRefreshToken ` is false and if the access token we have retrieved is already expired, in that case we dont require this information at this point.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/dao/AccessTokenDAOImpl.java
AfraHussaindeen: The access token in this scenario expired on its own, simply due to its predefined lifetime. Here’s a breakdown:

- Time-Based Expiration: 
Every access token is created with a specific validity period (e.g., 3600 seconds). The token becomes invalid as soon as the current time surpasses this timestamp. This is what happened in the scenario I described. The token simply ran out of time, independent of any other user or system action.

- Revocation 
This is the process you mentioned, where issuing a new token can invalidate an old one. During a refresh token grant flow , the system generates a new access token. While this also makes the old token invalid, it's a separate mechanism from the natural time-based expiration.


In the scenario I have mentioned in the issue, the token didn't expire because the user's profile was updated or because a token refresh occurred. It expired simply because its own expiry time reached. 

So, to answer your question, the token expired due to its own time-based validity, not as a result of the token refresh action.

piraveena (approved):

PR #3193, opened 2026-04-10
Clear OAuth cache using persisted token scopes on revoke

KD23243:
## Purpose
During token revocation, the in-memory `accessTokenDO` may have had its scopes mutated or filtered earlier in the request lifecycle. The existing cache-clearing path uses those (potentially mutated) scopes as part of the cache key, so entries originally cached under the persisted scopes can be left behind.

## Related Issue
N/A

## Implementation
- Added `OAuthUtil.clearOAuthCacheUsingPersistedScopes(...)` which:
  - Returns early if `OAuthServerConfiguration#getAllowedScopes` is empty or the access token is blank.
  - Fetches the access token from `OAuthTokenPersistenceFactory.getAccessTokenDAO().getAccessToken(token, true)` to obtain the original persisted scopes.
  - Builds the scope string from the persisted scopes and calls `clearOAuthCache` for both the `(consumerKey, authzUser, scope, tokenBindingReference)` and `(consumerKey, authzUser, scope)` key variants.
  - Catches `IdentityOAuth2Exception` from the DAO and logs it without propagating.
- Wired the new helper into `OAuth2Service#revokeTokenByOAuthClient`, called right after the existing `clearOAuthCache` invocations.
- Added unit tests in `OAuthUtilTest` covering:
  - Empty `allowedScopes` short-circuit.
  - Blank access token short-circuit.
  - DB token lookup returning `null`.
  - DB token returning `null` scope array.
  - DB token returning empty scope array.
  - Happy path: cache cleared with persisted scopes for both key variants.
  - DAO throwing `IdentityOAuth2Exception` is swallowed and no cache clear is invoked.

## Summary by CodeRabbit

* **New Features**
  * Enhanced OAuth token revocation with improved cache clearing mechanism to ensure complete invalidation of revoked tokens.

* **Tests**
  * Added comprehensive unit tests covering token revocation cache clearing scenarios, including edge cases and exception handling.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
@@ -50,6 +50,7 @@ import org.wso2.carbon.identity.oauth.cache.AuthorizationGrantCacheKey;
 import org.wso2.carbon.identity.oauth.cache.OAuthCache;
 import org.wso2.carbon.identity.oauth.cache.OAuthCacheKey;
 import org.wso2.carbon.identity.oauth.common.OAuthConstants;
+import org.wso2.carbon.identity.oauth.config.OAuthServerConfiguration;
 import org.wso2.carbon.identity.oauth.dao.OAuthAppDO;
 import org.wso2.carbon.identity.oauth.dto.OAuthConsumerAppDTO;
 import org.wso2.carbon.identity.oauth.event.OAuthEventInterceptor;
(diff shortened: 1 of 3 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthUtil.java
```diff
@@ -1577,4 +1579,49 @@ private static String getUserStoreDomainOfParentUser(String parentUserId, String
...
+        if (StringUtils.isBlank(accessToken)) {
+            return;
+        }
+
+        try {
+            // The in-memory scopes may be mutated during validation. To avoid cache-key
+            // mismatches, retrieve the original scopes from the database before clearing
+            // the OAuth cache.
+            AccessTokenDO dbTokenDO = OAuthTokenPersistenceFactory.getInstance()
+                    .getAccessTokenDAO()
```
Thumimku, line 1610: use token provider if exists

PR #3220, opened 2026-04-27
Unit test for fst cache versioning

Thumimku:

This covers unit tests for cache versioning related to the FST serializer. Currently, support is limited to the AuthorizationGrantCache and DeviceGrantCache.

The two test methods in CacheEntryVersioningTest.java:                                                                                                                                                   
                                                                                                                                                                                                           
  ---                                                                                                                                                                                                      
  testFieldVersioningContract (line 96)
                                                                                                                                                                                                           
  Enforces the full FST versioning contract on each cache entry class: 
  - Every field in the v0 baseline must still exist, in the same declaration order, and must not have @Version                                                                                             
  - Every field added after the baseline must have @Version(n) where n ≥ 1                                                                                                                                 
  - The total count of @Version-annotated fields must exactly match expectedVersionedCount                                                                                                                 
                                                                                                                                                                                                           
  This is the strict guardrail — it catches someone adding a new field without @Version, which would break FST deserialization of older cached objects.
                                                                                                                                                                                                           
  ---                                                                  
  testBaselineFieldOrderPreserved (line 102)                                                                                                                                                               
                                                                       
  A lighter sanity check — it only verifies that the baseline fields still exist in their original declaration order. It doesn't inspect @Version annotations or care about new fields.
                                                                                                                                                                                                           
  This catches destructive changes like renaming, reordering, or removing a field that was part of the v0 snapshot, which would break backward compatibility with already-serialized cache entries.        
                                                                                                                                                                                                           
  ---                                                                                                                                                                                                      
  Together: the first test guards forward additions (new fields must be versioned), and the second guards backward stability (existing baseline fields must not move or disappear).

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth/cache/CacheEntryVersioningTest.java
+++ b/components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth/cache/CacheEntryVersioningTest.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.carbon.identity.oauth.cache;
+
+import org.testng.annotations.DataProvider;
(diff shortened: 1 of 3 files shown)


components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth/cache/CacheEntryVersioningTest.java
```diff
@@ -0,0 +1,95 @@
...
+                    "isPreIssueIDTokenActionsExecuted",
+                    "preIssueIDTokenActionDTO"
+            ));
+
+    @DataProvider(name = "cacheEntries")
+    public Object[][] cacheEntries() {
+
+        return new Object[][] {
+                {
+                        AuthorizationGrantCacheEntry.class,
```
piraveena, line 84: Shall we incorporate DeviceAuthorizationGrantCache too with this?

Thumimku: added via https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/3220/commits/d4b1a032489e6fcdff109b0e55c816a0ff830c2c

piraveena (approved):

PR #3221, opened 2026-04-27
[next] Fix missing isk claim in ID token for Device Code App Native Auth flow

HasiniSama:
$subject

Issues: https://github.com/wso2/product-is/issues/27473

The isk claim was absent from ID tokens issued via the Device Code grant when App Native Authentication was enabled.

**Root cause:** The session context identifier was never propagated through the Device Code flow's cache pipeline.

1. `DeviceAuthorizationGrantCacheEntry` had no `sessionContextIdentifier` field, so it was dropped when user attributes were cached after device authorization.
2. `AccessTokenIssuer.getAuthzGrantCacheEntryFromDeviceCode()` did not copy `sessionContextIdentifier` when converting `DeviceAuthorizationGrantCacheEntry` to `AuthorizationGrantCacheEntry`.
3. The `SESSION_IDENTIFIER` property was never set on `OAuthTokenReqMessageContext` before ID token building, so DefaultIDTokenBuilder's fallback lookup also returned null on the first token request.

## Justification for adding the sessionContextIdentifier in the token request context:

- During the `/token` call we create the `AuthorizationGrantCacheEntry` from `DeviceAuthorizationGrantCache` [1]. We set the **sessioncontextidentifer** here too.

- However this `AuthorizationGrantCacheEntry` get stored in cache only after the ID token is issues. Hence the first call the cache entry is not there. 

- ID token is built from either this `AuthorizationGrantCacheEntry` or the **token request context** [2].

- So we can set the **sessioncontextidentifer** here to token request context to handle this first call. [3] 

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java#L820
[2]https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/master/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultIDTokenBuilder.java#L210
[3] https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/3221/changes#diff-a2c746fa4a16f522d111c7be3da6396a523c3b56e33d2936a66853de7ef67affR231

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth/pom.xml
@@ -62,11 +62,6 @@
             <groupId>commons-lang.wso2</groupId>
             <artifactId>commons-lang</artifactId>
         </dependency>
-        <dependency>
-            <groupId>de.ruedigermoeller</groupId>
-            <artifactId>fst</artifactId>
-            <scope>provided</scope>
-        </dependency>
          <dependency>
             <groupId>org.ops4j.pax.logging</groupId>
             <artifactId>pax-logging-api</artifactId>
--- a/components/org.wso2.carbon.identity.oauth/pom.xml
+++ b/components/org.wso2.carbon.identity.oauth/pom.xml
@@ -62,6 +62,11 @@
             <groupId>commons-lang.wso2</groupId>
             <artifactId>commons-lang</artifactId>
         </dependency>
+        <dependency>
+            <groupId>de.ruedigermoeller</groupId>
+            <artifactId>fst</artifactId>
+            <scope>provided</scope>
+        </dependency>
          <dependency>
             <groupId>org.ops4j.pax.logging</groupId>
             <artifactId>pax-logging-api</artifactId>
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/device/cache/DeviceAuthorizationGrantCacheEntry.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/device/cache/DeviceAuthorizationGrantCacheEntry.java
@@ -26,6 +26,14 @@ import java.util.Map;
 
 /**
  * Contains authenticated user attributes.
+ *
+ * <p><b>Serialization Compatibility Rules (FST):</b></p>
+ * <ul>
+ *   <li>New fields MUST be annotated with {@code @Version(n)} where n is incremented for each release.</li>
+ *   <li>New fields MUST be declared AFTER all existing fields.</li>
+ *   <li>NEVER remove, reorder, or change the type of existing fields.</li>
+ *   <li>Code must handle null/default values for versioned fields when deserializing old data.</li>
+ * </ul>
  */
 public class DeviceAuthorizationGrantCacheEntry extends CacheEntry {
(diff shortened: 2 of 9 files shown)


pom.xml
```diff
@@ -632,6 +632,13 @@
                 <artifactId>org.wso2.carbon.identity.multi.attribute.login.mgt</artifactId>
                 <version>${carbon.identity.framework.version}</version>
             </dependency>
+            <dependency>
+                <groupId>de.ruedigermoeller</groupId>
+                <artifactId>fst</artifactId>
+                <version>${de.ruedigermoeller.fst.version}</version>
```
piraveena: Do we need this dependency? We have already added this dependency for previous fixes

Thumimku: This is duplicate entry

components/org.wso2.carbon.identity.oauth/pom.xml
```diff
@@ -61,6 +61,11 @@
         <dependency>
             <groupId>commons-lang.wso2</groupId>
             <artifactId>commons-lang</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>de.ruedigermoeller</groupId>
```
Thumimku: this is duplicate entry, please check again

piraveena:
@coderabbitai review

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/device/cache/DeviceAuthorizationGrantCacheEntry.java
```diff
@@ -18,6 +18,7 @@
 
 package org.wso2.carbon.identity.oauth2.device.cache;
 
+import org.nustaq.serialization.annotations.Version;
```
Thumimku, line 21: add these rules in the class doc
```
 * <p><b>Serialization Compatibility Rules (FST):</b></p>
 * <ul>
 *   <li>New fields MUST be annotated with {@code @Version(n)} where n is incremented for each release.</li>
 *   <li>New fields MUST be declared AFTER all existing fields.</li>
 *   <li>NEVER remove, reorder, or change the type of existing fields.</li>
 *   <li>Code must handle null/default values for versioned fields when deserializing old data.</li>
 * </ul>
```

pom.xml
HasiniSama: This is now removed.

components/org.wso2.carbon.identity.oauth/pom.xml
HasiniSama: This is now removed.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/device/cache/DeviceAuthorizationGrantCacheEntry.java
HasiniSama: This is now updated.

piraveena:
@coderabbitai review

piraveena (approved):

PR #3236, opened 2026-05-13
Fix incorrect accessing org id returned for shared user direct logins through enhanced organization authentication

Yasasr1:
### Purpose
$subject

### Related issue
- https://github.com/wso2/product-is/issues/27371

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
@@ -357,8 +357,12 @@ public class AuthorizationCodeGrantHandler extends AbstractAuthorizationGrantHan
                try {
                    authenticatedUser.setUserResidentOrganization(organizationManager.resolveOrganizationId(
                            authenticatedUser.getTenantDomain()));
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/AuthorizationCodeGrantHandler.java
```diff
@@ -357,8 +357,12 @@ private void resolveSharedUserDetails(AuthenticatedUser authenticatedUser, Strin
                try {
                    authenticatedUser.setUserResidentOrganization(organizationManager.resolveOrganizationId(
                            authenticatedUser.getTenantDomain()));
-               authenticatedUser.setAccessingOrganization(organizationManager.resolveOrganizationId(
-                       IdentityTenantUtil.resolveTenantDomain()));
+                   String accessingOrganization = authorizationGrantCacheEntry.getAccessingOrganization();
+                   if (StringUtils.isBlank(accessingOrganization)) {
```
SujanSanjula96, line 361: Shall we try a scenario where refresh token call after user attribute update since authorization grant cache can be cleared in that scenario

Yasasr1: tried this flow without issues. this logic is not getting executed for refresh token calls

SujanSanjula96 (approved):

PR #3249, opened 2026-06-07
Add Error Constant for Secret not Found

hisanhunais:
### Purpose
- Add error constant for secret not found


hisanhunais:
Closing this as it is not needed now

PR #2437, opened 2024-04-26
Generic the capability to enrich JWT and ID_token with custom claims

Sachin-Mamoru:
#### Purpose:
Introduces a constant to identify dynamic token data functions, enhancing the security and customization of token handling.

#### Approach:
- New Constant Introduction: A constant DYNAMIC_TOKEN_DATA_FUNCTION is added to identify scripts related to token data dynamically, improving maintainability and readability.
- Code Simplification: The condition checking whether user consent is available before fetching token attributes is removed, thereby simplifying the flow and always ensuring extended token attributes are considered.
- Improved Token Attribute Handling: Modifies the condition to check for the presence of a dynamic token data function directly in the script content, ensuring more accurate and secure token handling.

PR #3259, opened 2026-06-16
Restrict federated assertion claims to configured access token attributes

### LINKED ISSUE #27611: Inconsistency on How Claims are Propagated to Access Token in JWT Bearer Grant
### Description

When the JWT Bearer grant is configured in Identity Server 7.2, the claims present in the `assertion` (i.e., the JWT issued by the trusted identity provider) are not propagated to the access token by default.

However, when **any claim** is configured under the **“Access Token Attributes”** section in the **Protocol** tab, the claims from the assertion are propagated into the access token.

This behavior occurs even if the selected claim is not directly related to the claims present in the `assertion`.

### Steps to Reproduce

Inspect the access token returned in the JWT Bearer grant under the following conditions,

1. No claims are configured under  **“Access Token Attributes”**.
2. Add a random claim to the  **“Access Token Attributes”**.

### Version

wso2is-7.2.0

### Environment Details (with versions)

_No response_

### Developer Checklist

- [ ] [Behavioural Change] Does this change introduce a behavioral change to the product?
- [ ]  ↳ Approved by team lead
- [ ]  ↳ Label `impact/behavioral-change` added
- [ ] [Migration Impact] Does this change have a migration impact?
- [ ]  ↳ Migration label added (e.g., `7.2.0-migration`)
- [ ]  ↳ Migration issues created and linked
- [ ] [New Configuration] Does this change introduce a new configuration?
- [ ]  ↳ Label `config` added
- [ ]  ↳ Configuration is properly documented

RushanNanayakkara:
## Purpose
In the JWT bearer grant, all assertion claims for a federated user are added to the JWT access token **without** applying the application's configured *Access Token Attributes* — so the configured claim list behaves as an on/off switch rather than an allow-list. See wso2/product-is#27611.

## Approach
Introduce `OAuth.OpenIDConnect.RestrictAssertionClaimsToConfiguredAttributes` (default **false**). When enabled, `JWTAccessTokenOIDCClaimsHandler` restricts the federated user's assertion claims to the application's configured access token attributes (returns a filtered copy). Default false preserves existing behaviour — fully backward compatible.

**Scope: the JWT access-token handler only.** The *Access Token Attributes* list is a claims-separation concept — it governs the access token only when access-token claims separation is enabled (which is exactly what routes the access token through `JWTAccessTokenOIDCClaimsHandler`). When claims separation is disabled, the access token (like the id_token) is governed by scopes, so there is no access-token-attribute contract to enforce; neither the access token nor the id_token is changed in that mode.

## Related Issues
- Fixes wso2/product-is#27611
- Config-mapping companion PR: wso2/carbon-identity-framework#8160

## Tests
- Unit: `JWTAccessTokenOIDCClaimsHandlerTest` — config off → unfiltered passthrough; config on → restricted to the configured attributes.
- Runtime: separation on + config on → access token restricted to the configured attributes; separation off → no change (config has no effect); id_token unchanged in every case.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/config/OAuthServerConfiguration.java
@@ -303,6 +303,9 @@ public class OAuthServerConfiguration {
 
     // Property added to preserve the backward compatibility to send the original claim uris comes in the assertion.
     private boolean convertOriginalClaimsFromAssertionsToOIDCDialect = false;
+    // When enabled, claims taken from the assertion for a federated user are restricted to the application's
+    // configured access token attributes. Default false to preserve backward compatibility.
+    private boolean restrictAssertionClaimsToConfiguredAttributes = false;
     // This property will decide whether to send only mapped roles received from the federated IdP
     private boolean returnOnlyMappedLocalRoles = false;
 
@@ -2173,6 +2176,10 @@ public class OAuthServerConfiguration {
         return convertOriginalClaimsFromAssertionsToOIDCDialect;
     }
 
+    public boolean isRestrictAssertionClaimsToConfiguredAttributes() {
+        return restrictAssertionClaimsToConfiguredAttributes;
+    }
+
     public boolean isReturnOnlyMappedLocalRoles() {
         return returnOnlyMappedLocalRoles;
     }
@@ -3990,6 +3997,14 @@ public class OAuthServerConfiguration {
                 convertOriginalClaimsFromAssertionsToOIDCDialect = Boolean
                         .parseBoolean(convertOriginalClaimsFromAssertionsToOIDCDialectElement.getText().trim());
             }
+
+            OMElement restrictAssertionClaimsToConfiguredAttributesElement = openIDConnectConfigElem
+                    .getFirstChildWithName(getQNameWithIdentityNS(
+                            ConfigElements.OPENID_CONNECT_RESTRICT_ASSERTION_CLAIMS_TO_CONFIGURED_ATTRIBUTES));
+            if (restrictAssertionClaimsToConfiguredAttributesElement != null) {
+                restrictAssertionClaimsToConfiguredAttributes = Boolean
+                        .parseBoolean(restrictAssertionClaimsToConfiguredAttributesElement.getText().trim());
+            }
             OMElement addUnmappedUserAttributesElement = openIDConnectConfigElem.getFirstChildWithName(
                     getQNameWithIdentityNS(ConfigElements.OPENID_CONNECT_ADD_UN_MAPPED_USER_ATTRIBUTES));
             if (addUnmappedUserAttributesElement != null) {
@@ -4692,6 +4707,8 @@ public class OAuthServerConfiguration {
                 "IDTokenCustomClaimsCallBackHandler";
         public static final String OPENID_CONNECT_CONVERT_ORIGINAL_CLAIMS_FROM_ASSERTIONS_TO_OIDCDIALECT =
                 "ConvertOriginalClaimsFromAssertionsToOIDCDialect";
+        public static final String OPENID_CONNECT_RESTRICT_ASSERTION_CLAIMS_TO_CONFIGURED_ATTRIBUTES =
+                "RestrictAssertionClaimsToConfiguredAttributes";
         // Property to decide whether to add tenant domain to id_token.
         private static final String OPENID_CONNECT_ADD_TENANT_DOMAIN_TO_ID_TOKEN = "AddTenantDomainToIdToken";
         // Property to decide whether to add userstore domain to id_token.
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/JWTAccessTokenOIDCClaimsHandler.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/JWTAccessTokenOIDCClaimsHandler.java
@@ -177,10 +177,28 @@ public class JWTAccessTokenOIDCClaimsHandler implements CustomClaimsCallbackHand
         Object hasNonOIDCClaimsProperty = requestMsgCtx.getProperty(OIDCConstants.HAS_NON_OIDC_CLAIMS);
         if (isPreserverClaimUrisInAssertion(requestMsgCtx) || (hasNonOIDCClaimsProperty != null
                 && (Boolean) hasNonOIDCClaimsProperty)) {
+            if (OAuthServerConfiguration.getInstance().isRestrictAssertionClaimsToConfiguredAttributes()) {
+                return filterRequestedClaims(userClaimsInOIDCDialect, allowedClaims);
+            }
             return userClaimsInOIDCDialect;
         } else {
             return filterClaims(userClaimsInOIDCDialect, requestMsgCtx);
         }
+    }
+
+    /**
+     * Restrict the federated user's assertion claims to the application's configured access token attributes.
+     *
+     * @param userClaimsInOIDCDialect User claims in OIDC dialect.
+     * @param allowedClaims           Configured access token attributes.
+     * @return Claims restricted to the configured access token attributes.
+     */
+    private Map<String, Object> filterRequestedClaims(Map<String, Object> userClaimsInOIDCDialect,
+                                                      List<String> allowedClaims) {
+
+        return userClaimsInOIDCDialect.entrySet().stream()
+                .filter(entry -> allowedClaims.contains(entry.getKey()))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
     }
 
     private void setOrganizationSwitchAttributes(OAuthTokenReqMessageContext requestMsgCtx)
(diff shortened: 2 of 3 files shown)


RushanNanayakkara:
Configuration companion PR: wso2/carbon-identity-framework#8160

RushanNanayakkara:
## Runtime verification (re-tested) — WSO2 IS 7.4.0-SNAPSHOT

**Scope:** the fix lives solely in `JWTAccessTokenOIDCClaimsHandler` (3 files — the `RestrictAssertionClaimsToConfiguredAttributes` config in `OAuthServerConfiguration`, the in-bypass filter in `JWTAccessTokenOIDCClaimsHandler`, and `JWTAccessTokenOIDCClaimsHandlerTest`). `DefaultOIDCClaimsCallbackHandler` and the id_token path are **not** modified.

The PR-head `org.wso2.carbon.identity.oauth` jar was rebuilt (JDK 21), repacked into the deployed bundle slot, and deployed into a `wso2is-7.4.0-SNAPSHOT` pack — loads clean (no `org.wso2.carbon.identity.oauth` bundle resolution errors). The **JWT bearer grant** was exercised end-to-end for a **federated user**.

**Setup**
- Federated assertion carries `email`, `given_name`, `family_name`, `custom_assertion_claim`. (`address` is never in the assertion — used to confirm a *configured-but-absent* attribute contributes nothing.)
- App `JWTBearer74` (v3), trusted token issuer `local-selfsigned-issuer`. The federated bypass is entered via the **`HAS_NON_OIDC_CLAIMS`** trigger (this pack runs `ConvertOriginalClaimsFromAssertionsToOIDCDialect = true`). The support/7.2 line (#2437) covers the **`isPreserverClaimUrisInAssertion`** trigger — between the two, both bypass triggers are exercised.
- Axes: **`EnableClaimsSeparationForAccessToken`** (selects the access-token handler — `JWTAccessTokenOIDCClaimsHandler` when ON) × **`RestrictAssertionClaimsToConfiguredAttributes`** (this PR's flag, default `false`).

### Correctness — separation × restrict

| separation | restrict | access-token claims | id_token claims |
|---|---|---|---|
| ON | OFF | `[]`→none; `[email]`/`[email,given_name]`/`[address]` → **all 4 leak** | all 4 |
| ON | **ON** | `[]`→none; `[email]`→**email**; `[email,given_name]`→**email+given_name**; `[address]`→**none** | all 4 (unchanged) |
| OFF | OFF | **all 4 leak** | all 4 |
| OFF | ON | **all 4 leak — no change** (separation off → the unmodified `DefaultOIDCClaimsCallbackHandler` builds the access token, so this flag is a no-op) | all 4 (unchanged) |

→ The access token is restricted to the application's configured access-token attributes **only under separation-on**; unconfigured assertion claims (`family_name`, `custom_assertion_claim`) are dropped. The **id_token is unchanged in every case** — confirming the access-token-only scoping.

### Backward compatibility — patched (flag OFF) vs pristine unpatched jar

Ran separation {ON, OFF} × `accessTokenAttributes` {`[]`, `[email]`, `[email,given_name]`, `[address]`} + id_token, first with the patched jar (flag OFF), then with the original **unpatched** jar — same pack and harness, only the jar swapped.

**Result: byte-for-byte identical token claim sets across all 12 comparison points.** With the flag off (the default) the patched build is indistinguishable from the unpatched product — no existing deployment changes unless it opts in.

### Unit tests
`JWTAccessTokenOIDCClaimsHandlerTest` — **Tests run: 11, Failures: 0** (JDK 21), including the two new cases:
- `testFederatedAssertionClaimsRestrictedWhenConfigEnabled` — restrict on → claims filtered to the configured attributes.
- `testFederatedAssertionClaimsNotRestrictedWhenConfigDisabled` — restrict off → unfiltered passthrough (backward compat).

> "all 4" = `email, given_name, family_name, custom_assertion_claim`.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultIDTokenBuilder.java
```diff
@@ -507,8 +507,17 @@ private CustomClaimDTO handleOIDCCustomClaims(OAuthTokenReqMessageContext tokReq
...
+        // access-token-only claim restriction to the id_token. Reset afterwards as the same request
+        // context is reused to issue the access token.
+        tokReqMsgCtx.addProperty(OIDCConstants.ID_TOKEN_ISSUANCE, true);
+        JWTClaimsSet customClaimsAddedJWTClaimSet;
+        try {
+            customClaimsAddedJWTClaimSet =
+                    claimsCallBackHandler.handleCustomClaims(jwtClaimsSetBuilderCopy, tokReqMsgCtx);
+        } finally {
+            tokReqMsgCtx.addProperty(OIDCConstants.ID_TOKEN_ISSUANCE, false);
+        }
```
shashimalcse: i dont think we need this change here

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultOIDCClaimsCallbackHandler.java
```diff
@@ -188,12 +191,58 @@ private Map<String, Object> getUserClaimsInOIDCDialect(OAuthTokenReqMessageConte
...
                 && (Boolean) hasNonOIDCClaimsProperty)) {
+            // When the access token claims separation is disabled this handler also builds the access token.
+            // In that case, restrict the federated assertion claims to the configured access token attributes
+            // if the config is enabled. The id_token is left unchanged.
+            if (OAuthServerConfiguration.getInstance().isRestrictAssertionClaimsToConfiguredAttributes()
+                    && !isIDTokenIssuance(requestMsgCtx)) {
+                String clientId = requestMsgCtx.getOauth2AccessTokenReqDTO().getClientId();
+                String spTenantDomain = requestMsgCtx.getOauth2AccessTokenReqDTO().getTenantDomain();
+                return filterRequestedClaims(userClaimsInOIDCDialect,
+                        getAccessTokenClaims(clientId, spTenantDomain));
```
shashimalcse: i dont think we need to filter based on allowed access token claims. check how filterOIDCClaims handle in there

RushanNanayakkara: Allowed access token claims not being honored is the reported issue (wso2/product-is#27611).
`filterOIDCClaims` filters by the requested OIDC scopes, not by the application's
configured access-token attributes. This PR's fix needs the configured access-token
attribute list (`getAccessTokenClaims`), which `filterOIDCClaims` doesn't consider.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultIDTokenBuilder.java
RushanNanayakkara: This is needed because the new restriction lives in `DefaultOIDCClaimsCallbackHandler`,
which is shared between two tokens:

- the **id_token** — always built through this `DefaultIDTokenBuilder` path, and
- the **access token** — when access-token claims separation is *disabled*, the access
  token is also built by this handler (`JWTTokenIssuer` →
  `ClaimHandlerUtil.getClaimsCallbackHandler()` returns `DefaultOIDCClaimsCallbackHandler`).

Both go through the same `getUserClaimsInOIDCDialect(OAuthTokenReqMessageContext)` where
`RestrictAssertionClaimsToConfiguredAttributes` is applied. The restriction must apply to
the **access token only** — the id_token must stay unchanged — but the handler has no way
to tell which token it's currently building. This marker is that signal: it's set during
id_token issuance so the handler skips the restriction for the id_token and applies it only
for the access token.

Without it, enabling the flag would also restrict the id_token's federated assertion claims
to the configured access-token attributes — a behaviour change to the id_token we explicitly
want to avoid. `DefaultOIDCClaimsCallbackHandlerTest.testFederatedAssertionClaimsNotRestrictedByDefaultHandlerForIdTokenWhenConfigEnabled`
covers exactly this case.

RushanNanayakkara:
## Scope finalized — `JWTAccessTokenOIDCClaimsHandler` only

An earlier iteration also extended the fix to `DefaultOIDCClaimsCallbackHandler` (the separation-off access token); that extension was **reverted**. The *Access Token Attributes* allow-list is a claims-separation concept — it only governs the access token when access-token claims separation is on, which is exactly when `JWTAccessTokenOIDCClaimsHandler` runs. With separation off the access token is scope-governed like the id_token, so there is no access-token-attribute contract to enforce and the fix does not apply there.

Final diff = **3 files**: the `RestrictAssertionClaimsToConfiguredAttributes` config in `OAuthServerConfiguration`, the in-bypass filter in `JWTAccessTokenOIDCClaimsHandler`, and `JWTAccessTokenOIDCClaimsHandlerTest`. `DefaultOIDCClaimsCallbackHandler`, `DefaultIDTokenBuilder`, and the id_token path are untouched.

The **Runtime verification** comment above reflects this final design and was re-tested end-to-end on both the 7.4 (master) and 7.2 (support) lines — correctness matrix, a byte-identical patched-vs-unpatched backward-compatibility check, and unit tests.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultOIDCClaimsCallbackHandler.java
RushanNanayakkara: As discussed, the changes were removed from DefaultOIDCClaimsCallbackHandler since with access and id token claim separation turned off, access token attributes should not be used for filtering access token attributes.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/openidconnect/DefaultIDTokenBuilder.java
RushanNanayakkara: As discussed, the changes were removed from DefaultOIDCClaimsCallbackHandler since with access and id token claim separation turned off, access token attributes should not be used for filtering access token attributes.

shashimalcse (approved):

PR #3262, opened 2026-06-22
Decode redirect_uris regexp string in DCR response when multiple URIs are registered

BimsaraBodaragama:
## Root cause and fix

- When an application is registered with multiple redirect URIs, WSO2 IS internally stores them as a single `regexp=(uri1|uri2|...)` callback string.
- The DCR `buildResponse` method returned this raw internal representation verbatim in the `redirect_uris` array of the POST/GET/PUT response, breaking the public DCR contract and APIM's round-trip.
- Fix: `buildResponse` now calls `buildRedirectUrisResponse`, which decodes the `regexp=(...)` form back into the original URI array when the config knob `OAuth.DCRM.DecodeRedirectUrisInResponse` is enabled. Default is `false` to preserve backward compatibility.

## Config knob

```toml
[oauth.dcrm]
decode_redirect_uris_in_response = true
```

Default: `false`. Requires server restart.

## Linked issue

https://github.com/wso2/product-is/issues/27851

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.dcr/src/main/java/org/wso2/carbon/identity/oauth/dcr/DCRMConstants.java
+++ b/components/org.wso2.carbon.identity.oauth.dcr/src/main/java/org/wso2/carbon/identity/oauth/dcr/DCRMConstants.java
@@ -103,6 +103,8 @@ public class DCRMConstants {
         public static final String UNAPPROVED_SOFTWARE_STATEMENT = "unapproved_software_statement";
     }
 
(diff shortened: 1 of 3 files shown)


BimsaraBodaragama:
Closing — wrong target. IS 7.3 fix goes to wso2-support branch, not public master.

BimsaraBodaragama:
The effort will be tracked via the below PR
 - https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/3263

PR #3277, opened 2026-07-14
Surface client errors as 4xx when agent application creation fails

RushanNanayakkara:

When a user-serving agent is created via `POST /scim2/Agents` with `IsUserServingAgent: true`, `UserApplicationCreationListener` creates an OAuth application for the agent. If that application creation fails with a client error — for example when the tenant's application limit is reached (`RLS-10001`) — the listener currently rethrows a generic `UserStoreException`. The SCIM layer cannot map that to a client error, so the API returns:

```
HTTP 500
{"detail":"Error in adding the user: A***... to the user store.","status":"500"}
```

This hides the actual cause from the client.

**Change:** in the catch block of `doPostAddUserWithID`, if the exception is an `IdentityApplicationManagementClientException` **with error code `RLS-10001` (resource limit reached)**, rethrow it as a `UserStoreClientException` carrying the original error code and message. Other client exceptions keep the existing behavior — the mapping is intentionally scoped to the limit-reached case only. `SCIMUserManager.createUser` resolves `UserStoreClientException` (directly or as the root cause of a wrapped `UserStoreException`) to a `BadRequestException`. The cause is intentionally not chained, because the SCIM layer resolves the root cause of the chain and it must resolve to the `UserStoreClientException`; the exception is logged before rethrowing.

The limit-reached client failure is logged at debug level (it is an expected, client-induced 4xx outcome and would otherwise flood the error log on repeated retries at the limit); all other failures keep error-level logging with the stack trace. Unit tests cover the limit-reached path (rollback + `UserStoreClientException` with code `RLS-10001`) and the scoping (other client error codes keep the existing `UserStoreException` behavior).

With this PR alone, the failure surfaces as a 400 through the SCIM layer's generic client-error handling. This is an intermediate state: wso2-extensions/identity-inbound-provisioning-scim2#799 builds on this error code and maps the case to its final form —

```
HTTP 403
{"scimType":"applicationLimitReached","detail":"Agent application creation failed. Maximum number of allowed applications have been reached.","status":"403"}
```

Both PRs should be shipped together; the end-to-end behavior change users observe is 500 → 403.

The rollback behavior (deleting the agent user when application creation fails) is unchanged.

Verified end-to-end on a local IS deployment with resource limiting enabled (together with the scim2 PR): at the application limit the API returns the 403 above, rollback still removes the agent, and both the plain-agent path and the user-serving-agent path with a free application slot still return 201.

Should be released together with wso2-extensions/identity-inbound-provisioning-scim2#799 so clients observe the final 403 response instead of the intermediate 400.

### Follow up actions

None.

## Developer Checklist (Mandatory)

- [ ] Complete the **Developer Checklist** in the related `product-is` issue to track any behavioral change or migration impact.

---

## Behavioral changes (explicit)

This PR changes the API response of `POST /scim2/Agents` **only** for the case where the agent is user-serving (`IsUserServingAgent: true`) and the automatic application creation fails with the resource-limit client error (`IdentityApplicationManagementClientException` with error code `RLS-10001`):

| | Before | After (with wso2-extensions/identity-inbound-provisioning-scim2#799) |
|---|---|---|
| HTTP status | `500` | `403` |
| `scimType` | – | `applicationLimitReached` |
| `detail` | `Error in adding the user: A***... to the user store.` (masked, unrelated to the real cause) | `Agent application creation failed. Maximum number of allowed applications have been reached.` |

(With this PR alone the case surfaces as `400` / `invalidValue` via the SCIM layer's generic client-error handling — an intermediate state; the two PRs should be shipped together.)

No behavioral change for:

- Agent creation succeeding, or failing for any other reason (server errors still return 500).
- Regular (non-agent) user creation.
- The rollback behavior — the agent user is still deleted when application creation fails.

Note: the Console handles the final `403` + `scimType: applicationLimitReached` response in wso2/identity-apps#10538.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/UserApplicationCreationListener.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/listener/UserApplicationCreationListener.java
@@ -66,6 +66,8 @@ public class UserApplicationCreationListener extends AbstractIdentityUserOperati
     private static final Log log = LogFactory.getLog(UserApplicationCreationListener.class);
     private static final String AGENT_LISTENER_ENABLE = "AgentIdentity.ApplicationCreatorListener.Enabled";
     private static final String AGENT_LISTENER_ORDER_ID = "AgentIdentity.ApplicationCreatorListener.Order";
+    // Error code set by the resource limit enforcement when application creation is blocked by the tier quota.
+    private static final String ERROR_CODE_RESOURCE_LIMIT_REACHED = "RLS-10001";
     boolean isEnabled = false;
 
     public UserApplicationCreationListener() {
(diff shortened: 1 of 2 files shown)


RushanNanayakkara:
Addressed the CodeRabbit review in 2ee3ff2820:

1. The resource-limit client failure is now logged at debug level (server faults keep error-level logging with the stack trace).
2. Added `UserApplicationCreationListenerTest` coverage for the new path: the limit-reached client exception surfaces as `UserStoreClientException` with code `RLS-10001` after rollback, and other client error codes keep the existing `UserStoreException` behavior.

sahandilshan (approved):

pavinduLakshan (approved):

PR #3297, opened 2026-08-21
Add diagnostic logs for refresh token validation failures

Miranlfk:
### Purpose

The diagnostic logs of a failed refresh token grant request stopped after the application validation, providing no information on why the request failed. Diagnostic logs are added for each refresh token validation failure so that the reason is explicit:

- The refresh token is invalid and no persisted token was found for it.
- The refresh token is not in a usable state, including the current state.
- The refresh token is not the latest token issued for the application.
- The refresh token is expired.

A log entry is also added when an authorization grant fails validation, so that the logs of a token request never end at the application validation regardless of the grant type in use.

Refresh tokens are never logged. To still provide context on the token used, the internal token id, the token state, the issued time, the validity period and the masked authorized user are logged. The SHA-256 hash of the refresh token is logged only when token logging is explicitly permitted through the REFRESH_TOKEN identity token configuration.

### DIFF
--- a/components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuthConstants.java
+++ b/components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuthConstants.java
@@ -854,6 +854,14 @@ public final class OAuthConstants {
             public static final String IMPERSONATOR = "impersonator";
             public static final String ACTOR = "actor";
             public static final String REQUESTED_AUTHORIZATION_DETAILS = "requested authorization details";
+            public static final String TOKEN_ID = "token id";
+            public static final String REFRESH_TOKEN_STATE = "refresh token state";
+            public static final String REFRESH_TOKEN_ISSUED_TIME = "refresh token issued time";
+            public static final String REFRESH_TOKEN_VALIDITY_PERIOD = "refresh token validity period (ms)";
+            /* Only the hash of a refresh token is ever logged and only when token logging is explicitly
+             permitted, since refresh tokens themselves cannot be logged. */
+            public static final String REFRESH_TOKEN_HASH = "refresh token hash";
+            public static final String ERROR_CODE = "error code";
         }
 
         /**
--- a/components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuthConstants.java
+++ b/components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuthConstants.java
@@ -858,8 +858,6 @@ public final class OAuthConstants {
             public static final String REFRESH_TOKEN_STATE = "refresh token state";
             public static final String REFRESH_TOKEN_ISSUED_TIME = "refresh token issued time";
             public static final String REFRESH_TOKEN_VALIDITY_PERIOD = "refresh token validity period (ms)";
-            /* Only the hash of a refresh token is ever logged and only when token logging is explicitly
-             permitted, since refresh tokens themselves cannot be logged. */
             public static final String REFRESH_TOKEN_HASH = "refresh token hash";
             public static final String ERROR_CODE = "error code";
         }
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/tokenprocessor/DefaultRefreshTokenGrantProcessor.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/tokenprocessor/DefaultRefreshTokenGrantProcessor.java
@@ -368,6 +368,17 @@ public class DefaultRefreshTokenGrantProcessor implements RefreshTokenGrantProce
             if (log.isDebugEnabled()) {
                 log.debug(String.format("Invalid Refresh Token provided for Client with Client Id : %s", clientId));
             }
+            if (LoggerUtils.isDiagnosticLogsEnabled()) {
+                LoggerUtils.triggerDiagnosticLogEvent(new DiagnosticLog.DiagnosticLogBuilder(
+                        OAuthConstants.LogConstants.OAUTH_INBOUND_SERVICE,
+                        OAuthConstants.LogConstants.ActionIDs.VALIDATE_REFRESH_TOKEN)
+                        .inputParam(LogConstants.InputKeys.CLIENT_ID, clientId)
+                        .resultMessage("The provided refresh token is invalid. No token was found for the given " +
+                                "refresh token and client id combination. The refresh token could have been " +
+                                "issued for a different application, already revoked or cleaned up after expiry.")
+                        .logDetailLevel(DiagnosticLog.LogDetailLevel.APPLICATION)
+                        .resultStatus(DiagnosticLog.ResultStatus.FAILED));
+            }
             throw new IdentityOAuth2Exception("Persisted access token data not found");
         }
         return true;
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/tokenprocessor/DefaultRefreshTokenGrantProcessor.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/tokenprocessor/DefaultRefreshTokenGrantProcessor.java
@@ -78,7 +78,7 @@ public class DefaultRefreshTokenGrantProcessor implements RefreshTokenGrantProce
         OAuth2AccessTokenReqDTO tokenReq = tokenReqMessageContext.getOauth2AccessTokenReqDTO();
         RefreshTokenValidationDataDO validationBean = OAuthTokenPersistenceFactory.getInstance().getTokenManagementDAO()
                 .validateRefreshToken(tokenReq.getClientId(), tokenReq.getRefreshToken());
-        validatePersistedAccessToken(validationBean, tokenReq.getClientId());
+        validatePersistedAccessToken(validationBean, tokenReq);
         validateReuseRefreshToken(validationBean, tokenReq.getClientId(), tokenReq.getTenantDomain());
         return validationBean;
     }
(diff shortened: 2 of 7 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -509,6 +509,23 @@ private OAuth2AccessTokenRespDTO validateGrantAndIssueToken(OAuth2AccessTokenReq
             if (log.isDebugEnabled()) {
                 log.debug("Invalid Grant provided by the client Id: " + tokenReqDTO.getClientId());
             }
+            /* The grant handlers log the specific reason for the failure. This log guarantees that the diagnostic
+             logs of a token request never end at the application validation, even if a grant handler could not
+             resolve a specific reason for the failure. The internal error message is deliberately not included.
+             It is not guaranteed to be free of user identifiers, and LoggerUtils.getSanitizedErrorMessage() can
+             only mask the value of the username request parameter, which grant types other than the password
+             grant do not carry. The error code carries the reason without carrying identifiers. */
```
sadilchamishka: We may not need this code level comment.

components/org.wso2.carbon.identity.oauth.common/src/main/java/org/wso2/carbon/identity/oauth/common/OAuthConstants.java
```diff
@@ -854,6 +854,14 @@ public static class InputKeys {
             public static final String IMPERSONATOR = "impersonator";
             public static final String ACTOR = "actor";
             public static final String REQUESTED_AUTHORIZATION_DETAILS = "requested authorization details";
+            public static final String TOKEN_ID = "token id";
+            public static final String REFRESH_TOKEN_STATE = "refresh token state";
+            public static final String REFRESH_TOKEN_ISSUED_TIME = "refresh token issued time";
+            public static final String REFRESH_TOKEN_VALIDITY_PERIOD = "refresh token validity period (ms)";
+            /* Only the hash of a refresh token is ever logged and only when token logging is explicitly
+             permitted, since refresh tokens themselves cannot be logged. */
```
sadilchamishka: These comments may not necessary.

sadilchamishka (dismissed):

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
```diff
@@ -138,6 +141,13 @@ public boolean validateGrant(OAuthTokenReqMessageContext tokReqMsgCtx)
                     ", Authorized User : " + validationBean.getAuthorizedUser() +
                     ", Token Scope : " + OAuth2Util.buildScopeString(validationBean.getScope()));
         }
+        if (LoggerUtils.isDiagnosticLogsEnabled()) {
```
sadilchamishka: The time based expiry of the refresh token is not checked here, it is checked in `issue()` via `isRefreshTokenExpired()`. So for an expired refresh token, which is the most common failure, this will first log `VALIDATE_REFRESH_TOKEN` with SUCCESS and then `issue()` will log `VALIDATE_REFRESH_TOKEN` again with FAILED for the same request. Having the same action id succeed and then fail makes the diagnostic trail confusing.

Shall we either use `ISSUE_ACCESS_TOKEN` as the action id for the failures logged from `issue()`, or move this success log to after the expiry check?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
```diff
@@ -138,6 +141,13 @@ public boolean validateGrant(OAuthTokenReqMessageContext tokReqMsgCtx)
...
+            DiagnosticLog.DiagnosticLogBuilder diagnosticLogBuilder = getRefreshTokenValidationLogBuilder(tokenReq,
+                    validationBean);
+            diagnosticLogBuilder.resultMessage("Refresh token validation is successful.")
```
sadilchamishka: This adds an APPLICATION level diagnostic log with the token id, state, issued time, validity period and the masked user for every successful refresh grant request. Is the success entry required, considering that this is one of the hottest OAuth paths and the token issuance already has its own diagnostic log? The gap this PR addresses is on the failure side.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
```diff
@@ -174,6 +184,17 @@ public OAuth2AccessTokenRespDTO issue(OAuthTokenReqMessageContext tokReqMsgCtx)
...
         if (validationBean == null || isRefreshTokenExpired(validationBean)) {
+            /* A null validation bean is not an expiry. It means the previous access token was never placed in the
+             message context, therefore the reason is reported separately to avoid pointing the investigation at
+             the refresh token itself. */
+            if (validationBean == null) {
+                logRefreshTokenValidationFailure("The validation data of the refresh token is not available in the " +
+                        "context of the token request, hence the refresh token could not be validated.", tokenReq,
+                        null);
+            } else {
+                logRefreshTokenValidationFailure("The provided refresh token is expired. A new refresh token needs " +
```
sadilchamishka, line 188: There is another path in `issue()` that returns the same `"Refresh token is expired."` error without a diagnostic log, the `IllegalArgumentException` handling for `OAuth2Util.ACCESS_TOKEN_IS_NOT_ACTIVE_ERROR_MESSAGE` from `createAccessTokenBean()`. That failure still ends the diagnostic trail at the application validation, which is the gap this PR is closing. Can we cover it as well?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
```diff
@@ -509,6 +509,23 @@ private OAuth2AccessTokenRespDTO validateGrantAndIssueToken(OAuth2AccessTokenReq
...
+            if (LoggerUtils.isDiagnosticLogsEnabled()) {
+                LoggerUtils.triggerDiagnosticLogEvent(new DiagnosticLog.DiagnosticLogBuilder(
+                        OAuthConstants.LogConstants.OAUTH_INBOUND_SERVICE,
+                        OAuthConstants.LogConstants.ActionIDs.ISSUE_ACCESS_TOKEN)
+                        .inputParam(LogConstants.InputKeys.CLIENT_ID, tokenReqDTO.getClientId())
+                        .inputParam(OAuthConstants.LogConstants.InputKeys.GRANT_TYPE, grantType)
+                        .inputParam(OAuthConstants.LogConstants.InputKeys.ERROR_CODE, errorCode)
+                        .resultMessage("Authorization grant validation failed for the provided grant type.")
+                        .logDetailLevel(DiagnosticLog.LogDetailLevel.APPLICATION)
```
sadilchamishka, line 521: The `authorizeAccessDelegation()` failure right below this block returns `UNAUTHORIZED_CLIENT` without any diagnostic log. If the intention is that the logs of a token request never end at the application validation, that branch needs the same treatment.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/tokenprocessor/DefaultRefreshTokenGrantProcessor.java
```diff
@@ -368,6 +368,17 @@ private boolean validatePersistedAccessToken(RefreshTokenValidationDataDO valida
             if (log.isDebugEnabled()) {
                 log.debug(String.format("Invalid Refresh Token provided for Client with Client Id : %s", clientId));
             }
+            if (LoggerUtils.isDiagnosticLogsEnabled()) {
```
sadilchamishka, line 373: This log carries only the client id, while all the other refresh token logs added in this PR are built through `getRefreshTokenValidationLogBuilder()` and carry the tenant domain as well. Shall we add the tenant domain here for consistency?

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandlerTest.java
```diff
@@ -119,9 +129,14 @@ public class RefreshGrantHandlerTest {
     private OAuthCache mockOAuthCache;
     private AuthorizationGrantCache mockAuthorizationGrantCache;
     private OAuthAppDO oAuthAppDO;
+    private MockedStatic<LoggerUtils> loggerUtils;
 
     @BeforeMethod
     public void init() {
+        /* The refresh token validation writes diagnostic logs. Diagnostic logging is mocked as enabled so that the
+         diagnostic log building code is exercised without requiring a carbon context to resolve the tenant. */
+        loggerUtils = mockStatic(LoggerUtils.class);
```
sadilchamishka: Mocking `LoggerUtils` statically in `@BeforeMethod` changes the behaviour of every existing test in this class, not only the new ones. Apart from enabling diagnostic logs for all of them, every other `LoggerUtils` method is now stubbed to its default too, for example `getSanitizedErrorMessage()` would return null. Would it be cleaner to open this mock only within the tests that need it?

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth/tokenprocessor/DefaultRefreshTokenGrantProcessorTest.java
```diff
@@ -131,7 +132,9 @@ public void setUp() throws Exception {
         identityUtilMockedStatic = mockStatic(IdentityUtil.class);
         serviceComponentHolderMockedStatic = mockStatic(OAuth2ServiceComponentHolder.class);
         loggerUtilsMockedStatic = mockStatic(LoggerUtils.class);
```
sadilchamishka, line 135: Same as in `RefreshGrantHandlerTest`, flipping this to `true` in the common setup changes the executed path of all the pre existing tests in this class. Enabling it inside the new test would keep the change contained.

components/org.wso2.carbon.identity.oauth/src/test/java/org/wso2/carbon/identity/oauth/tokenprocessor/DefaultRefreshTokenGrantProcessorTest.java
```diff
@@ -200,6 +203,35 @@ public void testValidateRefreshToken_gracefulDisabled_noStateFlipNoException() t
...
+        try {
+            processor.validateRefreshToken(buildTokenReqContext(CLIENT_ID, REFRESH_TOKEN, TENANT_DOMAIN));
+            fail("Expected IdentityOAuth2Exception when no persisted access token is found");
+        } catch (IdentityOAuth2Exception e) {
+            assertEquals(e.getMessage(), "Persisted access token data not found");
+        }
+
+        ArgumentCaptor<DiagnosticLog.DiagnosticLogBuilder> captor =
+                ArgumentCaptor.forClass(DiagnosticLog.DiagnosticLogBuilder.class);
+        loggerUtilsMockedStatic.verify(() -> LoggerUtils.triggerDiagnosticLogEvent(captor.capture()));
```
sadilchamishka: This `verify()` without a `times()` asserts exactly one `triggerDiagnosticLogEvent()` invocation in the flow, so the test will break if any other diagnostic log is added upstream of this. The helper in `RefreshGrantHandlerTest` uses `atLeastOnce()` with the last captured value, can we follow the same here?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/handlers/grant/RefreshGrantHandler.java
Miranlfk: That branch now calls logRefreshTokenValidationFailure() before returning. The message deliberately differs from the expiry one, it states that the access token issued against the refresh token is no longer active, so the two paths returning the identical "Refresh token is expired." error can be told apart in the logs.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/tokenprocessor/DefaultRefreshTokenGrantProcessor.java
Miranlfk: Added

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/token/AccessTokenIssuer.java
Miranlfk: Adjusted

sadilchamishka (approved):

PR #3302, opened 2026-08-27
Fix sub organization OIDC logout id token signature validation

### LINKED ISSUE #28231: `/oidc/` registered under the wrong rewrite category: sub-org logout on `/t/<tenant>/o/<org-id>/` executes in carbon.super and cannot resolve the application
### Description

For a **B2B sub-organization** OIDC application, **RP-initiated logout fails when invoked on the tenant-qualified endpoint form** `/t/<tenant-domain>/o/<org-id>/oidc/logout`, which is the form advertised by the sub-organization's OIDC discovery document and shown in the application's **Info** tab.

The request returns `access_denied` and the server log reports an ID-token decryption/signature failure. We observed **two different underlying failures** on this path, both of which are masked by that same misleading message:

1. The sub-organization's OAuth application cannot be resolved at all (`application.not.found`), because the request is not evaluated in the sub-organization's tenant context.
2. The application *is* resolved, and signature validation then fails, because the `id_token` was signed with the root organization's key while validation resolves the sub-organization's tenant.

Both were captured with the application's **Token Issuer set to the root organization**, which is a supported configuration. In both cases the reported error refers to JWE decryption, which does not reflect the actual cause and sends investigation in the wrong direction.

**Note on when this surfaces:** `OIDCLogoutServlet` only validates the `id_token_hint` when it receives **no** session cookie and falls back to `handleMissingSessionStateGracefully` (`HandleAlreadyLoggedOutSessionsGracefully` defaults to `true`). With a live session that path is not exercised, so the defect is masked. It reproduces reliably whenever the logout request reaches IS without a session cookie, for example a federated relying party on a different host, or a cleared or overwritten session cookie.

### Root cause

`/oidc/` is registered under the wrong rewrite category in `identity.xml`.

It is currently configured under `<OrgContextsToRewriteInTenantPerspective>`. Requests routed through that branch only get `accessingOrganizationId` / `applicationResidentOrganizationId` set on the `CarbonContext`; the **tenant context itself is not switched**. As a result, a request to `/t/carbon.super/o/<org-id>/oidc/logout` still executes in the `carbon.super` tenant, so `OAuth2Util.getAppInformationByClientId()` cannot locate the OAuth application that belongs to the sub-organization, and client validation fails.

The product already has the correct pattern in the other list: `<OrgResourceContextsToRewriteInTenantPerspective>` routes through `resolveAndValidateTenantQualifiedOrg`, which **does** switch the effective tenant to the sub-organization, but in 7.3.0 that list contains only `/identity/oidc/slo`.

The resulting `InvalidOAuthClientException` is then caught and reported by the servlet as an ID-token decryption (JWE) error, which is inaccurate: the token is a JWS, and the failure is application resolution rather than decryption.

Separately, once the application does resolve, `getTenantDomainForSignatureValidation` returns the **application owner's** tenant (the sub-organization) when `SignJWTWithSPKey=true` (the default). If the application's Token Issuer is the root organization, the `id_token` is signed with the root key, so signature validation still fails. This is the second failure listed above, and it means correcting the tenant switch alone is not sufficient for a root-issuer application.

### Affected code

- `repository/conf/identity/identity.xml`: `/oidc/` listed under `<OrgContextsToRewriteInTenantPerspective>` instead of `<OrgResourceContextsToRewriteInTenantPerspective>`
- `org.wso2.carbon.identity.context.rewrite.valve.TenantContextRewriteValve#invoke` (`identity-carbon-auth-rest` 1.12.11): the `OrgContextsToRewriteInTenantPerspective` branch sets the accessing organization but does not switch the tenant:
  ```java
  PrivilegedCarbonContext.getThreadLocalCarbonContext().setApplicationResidentOrganizationId(accessingOrgId);
  PrivilegedCarbonContext.getThreadLocalCarbonContext().setAccessingOrganizationId(accessingOrgId);
  // Carbon tenant remains carbon.super (from the /t/carbon.super/ prefix)
  ```
- `org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet` (`identity-inbound-auth-oauth` 7.4.99): `doGet` → `handleMissingSessionStateGracefully` → `getClientIdFromIdToken` → `validateIdToken` → `getTenantDomainForSignatureValidation`, where both the application lookup and the signature-validation tenant resolution take place

### Stacktrace

Both traces below are from IS 7.3.0 (update level 4), on a logout request to `/t/<tenant>/o/<org-id>/oidc/logout` with no session cookie, and in both cases the application's Token Issuer was the root organization. Note `TID: [-1234]` in both, showing the request executing as the **super tenant** rather than the sub-organization, which is the observable symptom of the tenant context not being switched.

**Variant 1: the sub-organization application cannot be resolved (`application.not.found`).**

```
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - opbs cookie not received. Missing session state.
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - 'SignJWTWithSPKey' property is set to : true
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - Error occurred while getting tenant domain for signature validation with id token: eyJ4NXQjUzI1NiI6...<id_token truncated>
org.wso2.carbon.identity.oauth.common.exception.InvalidOAuthClientException: application.not.found
    at org.wso2.carbon.identity.oauth.dao.OAuthAppDAO.handleRequestForANonExistingConsumerKey(OAuthAppDAO.java:2251)
    at org.wso2.carbon.identity.oauth.dao.OAuthAppDAO.getAppInformation(OAuthAppDAO.java:522)
    at org.wso2.carbon.identity.oauth2.util.OAuth2Util.getAppInformationByClientId(OAuth2Util.java:2619)
    at org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet.getTenantDomainForSignatureValidation(OIDCLogoutServlet.java:498)
    at org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet.validateIdToken(OIDCLogoutServlet.java:456)
    at org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet.getClientIdFromIdToken(OIDCLogoutServlet.java:1211)
    at org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet.handleMissingSessionStateGracefully(OIDCLogoutServlet.java:1103)
    at org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet.doGet(OIDCLogoutServlet.java:188)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:529)
    ...
    at org.wso2.carbon.tomcat.ext.servlet.DelegationServlet.service(DelegationServlet.java:68)
    at org.apache.catalina.core.ApplicationDispatcher.forward(ApplicationDispatcher.java:283)
    at org.wso2.carbon.identity.context.rewrite.valve.TenantContextRewriteValve.invoke(TenantContextRewriteValve.java:288)
    at org.wso2.carbon.identity.context.rewrite.valve.OrganizationContextRewriteValve.invoke(OrganizationContextRewriteValve.java:123)
    at org.wso2.carbon.identity.core.context.valve.IdentityContextCreatorValve.invoke(IdentityContextCreatorValve.java:55)
    at org.wso2.carbon.tomcat.ext.valves.CarbonContextCreatorValve.invoke(CarbonContextCreatorValve.java:63)
    ...
```

**Variant 2: the application is resolved, and signature validation then fails.** Here the validation tenant is correctly the sub-organization, but the `id_token` was signed with the root key:

```
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - OpBrowserState cookie not found in the inbound request. Attempting to extract the opBrowserState from cache.
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - Unable to resolve the opBrowser state.
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - opbs cookie not received. Missing session state.
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - 'SignJWTWithSPKey' property is set to : true
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - JWT signature will be validated with the service provider's tenant domain : <sub-org-tenant-domain>
TID: [-1234] [] [<timestamp>] [<correlation-id>] DEBUG {org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet} - Error occurred while decrypting the id token (JWE).org.wso2.carbon.identity.oauth2.IdentityOAuth2Exception: ID token signature validation failed.
    at org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet.getClientIdFromIdToken(OIDCLogoutServlet.java:1212)
    at org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet.handleMissingSessionStateGracefully(OIDCLogoutServlet.java:1103)
    at org.wso2.carbon.identity.oidc.session.servlet.OIDCLogoutServlet.doGet(OIDCLogoutServlet.java:188)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:529)
    ...
    at org.apache.catalina.core.ApplicationDispatcher.forward(ApplicationDispatcher.java:283)
    at org.wso2.carbon.identity.context.rewrite.valve.TenantContextRewriteValve.invoke(TenantContextRewriteValve.java:288)
    at org.wso2.carbon.identity.context.rewrite.valve.OrganizationContextRewriteValve.invoke(OrganizationContextRewriteValve.java:123)
    at org.wso2.carbon.identity.core.context.valve.IdentityContextCreatorValve.invoke(IdentityContextCreatorValve.java:55)
    at org.wso2.carbon.tomcat.ext.valves.CarbonContextCreatorValve.invoke(CarbonContextCreatorValve.java:63)
    ...
```

In both variants the failure is reported as `Error occurred while decrypting the id token (JWE)` / `ID token signature validation failed`, which does not reflect the actual cause in either case.

### Steps to Reproduce

1. Use a WSO2 IS 7.3.0 pack with organization management enabled (default).
2. Create a **sub-organization** under the root organization (`POST /api/server/v1/organizations`). Note its `<org-id>`.
3. Register an **OIDC application inside the sub-organization** (`POST /o/api/server/v1/applications`) with the `authorization_code` grant and a registered callback URL. Note its `<client-id>` and secret.
4. Create a **user in the sub-organization** (`POST /o/scim2/Users`).
5. Obtain an `id_token` for that application using the **tenant-qualified** endpoints, which are the form advertised by discovery:
   - authorize: `GET /t/carbon.super/o/<org-id>/oauth2/authorize?response_type=code&client_id=<client-id>&redirect_uri=<callback>&scope=openid`
   - token: `POST /t/carbon.super/o/<org-id>/oauth2/token` (authorization_code)
6. Invoke RP-initiated logout on the tenant-qualified form, **without sending a session cookie**, so that the `id_token_hint` validation path is exercised:
   ```
   GET /t/carbon.super/o/<org-id>/oidc/logout?id_token_hint=<id_token>&post_logout_redirect_uri=<registered-uri>
   ```
7. Repeat steps 5 and 6 with the application's **Token Issuer** set to the root organization and to the sub-organization, since both are supported configurations.

**Expected behaviour:** the logout completes and the browser is redirected to `post_logout_redirect_uri`, for both Token Issuer configurations, since `/t/<tenant>/o/<org-id>/…` is the recommended endpoint form for sub-organization OIDC applications.

**Actual behaviour:** the request fails with `access_denied`, and the server log reports `Error occurred while decrypting the id token (JWE).` followed by `ID token signature validation failed.` The real cause is either `InvalidOAuthClientException` (`application.not.found`), because the request executes in the `carbon.super` tenant, or a signature mismatch once the application resolves, because the token was signed with the root key while validation uses the sub-organization's tenant.

**Diagnostic note:** the same logout request against `/o/<org-id>/oidc/logout` does not fail, because that path resolves the tenant from the organization ID. That path is an internal redirection path and is not intended for external relying-party configuration, so it is noted here only as evidence that the failure is context resolution rather than the request itself.

### Suggested Fix

**Agreed scope:** logout should work correctly on the recommended `/t/carbon.super/o/<org-id>/…` endpoint form for **both** Token Issuer configurations, root organization and sub-organization. Supporting the `/o/<org-id>/…` form is explicitly out of scope, as it is intended for internal redirections.

1. Register `/oidc/` under `<OrgResourceContextsToRewriteInTenantPerspective>` so that it is routed through `resolveAndValidateTenantQualifiedOrg` and the effective tenant is switched to the sub-organization, matching the existing handling of `/identity/oidc/slo`.
2. Make signature validation resolve the same tenant that issued the token, derived from the application's Token Issuer, rather than the application owner's tenant. Correcting the tenant switch alone leaves the root-issuer case failing, as shown in Variant 2 above, where the application resolved correctly and validation still failed against the sub-organization's key.
3. Correct the misleading error message so that an application-resolution or client-validation failure is reported as such, instead of `Error occurred while decrypting the id token (JWE)`. The token is a JWS and no decryption is involved.

We only exercised `/oidc/logout` in this investigation, so we have not verified whether the other front-channel contexts (`/oauth2/token|authorize|userinfo|jwks`) show the same inconsistency under the tenant-qualified form.

### Please select the area issue is related to

API Access Mgt & Authorization

### Version

7.3.0 (reproduced on update level 4). Not verified on 7.1.0 or 7.2.0, though the same rewrite-category configuration is present in those lines and is likely affected.

### Environment Details (with versions)

Reproduced locally on WSO2 IS 7.3.0 update level 4, JDK 21 (Amazon Corretto), default H2 datasource, single node. Originally reported in a deployment where WSO2 API Manager 4.6.0 acts as a federated OIDC relying party in front of the sub-organization, but API Manager is not required to reproduce, as the minimal reproduction above uses IS alone.

### Developer Checklist

- [ ] [Behavioural Change] Does this change introduce a behavioral change to the product?
- [ ]  ↳ Approved by team lead
- [ ]  ↳ Label `impact/behavioral-change` added
- [ ] [Migration Impact] Does this change have a migration impact?
- [ ]  ↳ Migration label added (e.g., `7.2.0-migration`)
- [ ]  ↳ Migration issues created and linked
- [ ] [New Configuration] Does this change introduce a new configuration?
- [ ]  ↳ Label `config` added
- [ ]  ↳ Configuration is properly documented

RushanNanayakkara:
## Purpose

Resolves https://github.com/wso2/product-is/issues/28231

OIDC RP initiated logout fails with `access_denied` for an application in a sub organization, and the server log reports it as a JWE decryption failure, which is a misdiagnosis. The real failure is id token signature validation.

This is the out of the box configuration: a newly created sub organization application defaults its token issuer to the root organization.

## Goals

Validate the `id_token_hint` with the key that actually signed it when logout is called on `/t/<tenant-domain>/o/<organization-id>/`, and find the application reliably rather than only when a cache happens to be warm.

## Approach

Two separate causes.

**The wrong key.** The tenant qualified organization endpoints are served in the **root** tenant, so the id token is root signed. Validation resolved the **application owner** tenant, the sub organization, and used its key.

**The application was often not found at all.** The lookup used the deprecated `getAppInformationByClientId(clientId)`, which reads a tenant unscoped cache and otherwise falls back to the tenant in the thread local context. In the root tenant a sub organization application does not exist there, so the lookup succeeded only while the application happened to be in `AppInfoCache`, and failed with `application.not.found` otherwise. That is why one request produced two different errors, and why the failure depended on whether the cache was warm.

Two measured facts shape the fix.

Only one endpoint form sets an accessing organization:

| logout called at | accessing organization set |
|---|---|
| `/oauth2/...` | no |
| `/t/<tenant-domain>/oauth2/...` | no |
| `/o/<organization-id>/oauth2/...` | no |
| `/t/<tenant-domain>/o/<organization-id>/oauth2/...` | **yes** |

And the token issuer configured on an application only takes effect for tokens obtained through `/t/<tenant-domain>/o/<organization-id>/`. On the other forms it is ignored and the sub organization key signs. So the application configuration alone does not determine which key signed a given token.

`resolveSigningTenantDomain` therefore has two paths, decided by the accessing organization.

**No accessing organization.** The request is already served in the tenant that owns the application, and the token was issued in that same tenant, so the application owner tenant key is used.

**Accessing organization present.** The application is looked up with `getAppInformationByClientId(clientId, accessingTenantDomain)`, so it is found regardless of cache state. The signature is then validated with the key of the tenant that issues id tokens for that application: its configured token issuer, or, for an application created before that setting existed, the root organization of its own organization.

Two smaller fixes: the complete id token is no longer written to the logs, since it comes from a request parameter and is credential like (CWE-532); and the error now tells a JWE decryption failure apart from a signature failure using the `INVALID_ID_TOKEN` sub error code, rather than labelling every case a JWE failure.

## Behaviour change

**A relying party must use the same endpoint form for login and for logout.** A token obtained through `/o/<organization-id>/` or `/t/<org-handle>/` and then logged out through `/t/<tenant-domain>/o/<organization-id>/` is denied. Four such combinations work on the unpatched product.

Measured over every login form crossed with every logout form, for both token issuer values:

| application | token obtained at | logout called at | unpatched | with this change |
|---|---|---|---|---|
| sub organization, issuer = root organization | `/t/<tenant>/o/<org-id>/` | `/t/<tenant>/o/<org-id>/` | fails | **works** |
| no token issuer configured | `/t/<tenant>/o/<org-id>/` | `/t/<tenant>/o/<org-id>/` | fails | **works** |
| sub organization, issuer = root organization | `/t/<org-handle>/` | `/t/<tenant>/o/<org-id>/` | works | **fails** |
| sub organization, issuer = root organization | `/o/<org-id>/` | `/t/<tenant>/o/<org-id>/` | works | **fails** |
| no token issuer configured | `/t/<org-handle>/` | `/t/<tenant>/o/<org-id>/` | works | **fails** |
| no token issuer configured | `/o/<org-id>/` | `/t/<tenant>/o/<org-id>/` | works | **fails** |

The two that begin working are the reported defect and its equivalent for an application with no token issuer configured. The four that stop working all obtain the token on one endpoint form and call logout on another. Applications owned by the root organization are unaffected.

One qualification worth stating plainly: cross form logout is not refused uniformly. It still succeeds where the tenant the application issues from also owns the signing key, and fails otherwise. The same form expectation is a convention, not something the code enforces.

This behaviour change was agreed with the B2B team.

## Release note

Fixed OIDC RP initiated logout failing with "ID token signature validation failed" for a sub organization application accessed through the tenant qualified endpoint form. Relying parties must use the same endpoint form for login and for logout.

## Automation tests

**Unit tests.** Cases on `resolveSigningTenantDomain` cover the configured token issuer tenant, an application with no token issuer configured falling back to its root organization, requests with no accessing organization, the issuer claim not being consulted, and the cases that resolve to no tenant at all: an unparseable token, a client id that is not registered, a failure resolving the accessing organization, and organization management not being available.

On a recent JDK the suite needs `-Djacoco.skip=true -DargLine="-Dnet.bytebuddy.experimental=true"`. JaCoCo cannot instrument class file major version 69 and Mockito needs the ByteBuddy flag. This is a pre existing environment issue.

**Functional matrix, IS 7.3.0.** 39 cells: every login form crossed with every logout form, for both token issuer values, over a sub organization application, a root owned application with the root organization as the accessing organization, and an application with no token issuer stored. **33 of 39 pass unpatched, 31 of 39 with this change**: 2 fixed, 4 dropped by the behaviour change above.

**Behavioural suite, IS 7.3.0.** 92 assertions over two sub organizations, a root owned application, an application shared from the root organization, an organization addressed by a named handle, and a second application per organization. **12 assertions differ from the unpatched product: 6 fixed, 4 dropped by the behaviour change, and 2 that stay denied with a clearer reason.**

**Cold start, IS 7.3.0.** Logout as the first request after a restart, so the application lookup cannot be served from cache. This is where the `application.not.found` half of the defect appears. A configured application goes from 3 of 4 to 3 of 4, one same form case fixed and one cross form case dropped. An application with no token issuer stored goes from **2 of 5 to 3 of 5**.

**Public branch build.** The same behavioural suite (48 assertions) and cold suite (12 cells) run on a build of the public branch, with the cross form assertions differing as above. On that build a newly created sub organization application also defaults to the root organization as its token issuer, so the failure occurs without changing any setting.

**Negatives, all unchanged:** tampered signature, `alg=none`, an unparseable token, tokens signed by another organization in both directions, an issuer rewritten to name another tenant, an issuer origin swapped to a foreign host, another application's redirect uri, and an unregistered redirect uri. Adjacent endpoints are unchanged too: introspection, the refresh token grant, revocation, discovery and jwks.

## Security checks

- Followed secure coding standards? yes
- Ran FindSecurityBugs plugin and verified report? no. SpotBugs runs in the build; the report was not separately reviewed.
- Confirmed no keys, passwords, tokens, usernames or secrets committed? yes

The key used is always the one belonging to the tenant the application is configured to issue from, bounded to the application's own organization or its root. That is the same bound the product already enforces when the token issuer is selected (`getAllowedIssuerDetails` returns `Arrays.asList(primaryOrgId, orgId)`). A token this server did not issue is refused by the signature check.

## Test environment

WSO2 Identity Server 7.3.0 and a build of the public branch, Temurin JDK 25, macOS 15 arm64, H2.

### DIFF
--- a/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
+++ b/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
@@ -77,6 +77,7 @@ import org.wso2.carbon.identity.oidc.session.internal.OIDCSessionManagementCompo
 import org.wso2.carbon.identity.oidc.session.model.APIError;
 import org.wso2.carbon.identity.oidc.session.model.LogoutContext;
 import org.wso2.carbon.identity.oidc.session.util.OIDCSessionManagementUtil;
+import org.wso2.carbon.identity.organization.management.service.OrganizationManager;
 import org.wso2.carbon.identity.organization.management.service.exception.OrganizationManagementException;
 import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
 
@@ -556,6 +557,22 @@ public class OIDCLogoutServlet extends HttpServlet {
                     && tokenIssuer.equals(OAuth2OIDCConfigOrgUsageScopeUtils
                             .getIssuerLocation(issuerDetails.getIssuerTenantDomain()))) {
                 return issuerDetails.getIssuerTenantDomain();
+            }
+            /*
+             An application that predates the token issuer setting has no issuer configured, which is the
+             state of applications migrated from before the setting existed. Issuance then falls back to the
+             tenant serving the request, so the root organization of this application's own organization is
+             also a legitimate issuer for it. That is the same pair of tenants the setting itself allows.
+            */
+            if (issuerDetails == null || StringUtils.isEmpty(issuerDetails.getIssuerTenantDomain())) {
+                OrganizationManager organizationManager = OIDCSessionManagementComponentServiceHolder
+                        .getInstance().getOrganizationManager();
+                String primaryTenantDomain = organizationManager.resolveTenantDomain(organizationManager
+                        .getPrimaryOrganizationId(organizationManager.resolveOrganizationId(appTenantDomain)));
+                if (tokenIssuer.equals(OAuth2OIDCConfigOrgUsageScopeUtils
+                        .getIssuerLocation(primaryTenantDomain))) {
+                    return primaryTenantDomain;
+                }
             }
             /*
              The qualified issuer forms, /t/<tenant-domain>/oauth2/token and /o/<organization-id>/oauth2/token,
--- a/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
+++ b/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
@@ -543,7 +543,15 @@ public class OIDCLogoutServlet extends HttpServlet {
                         .getOrganizationManager().resolveTenantDomain(accessingOrgId);
                 oAuthAppDO = OAuth2Util.getAppInformationByClientId(clientId, accessingTenantDomain);
             } else {
+                /*
+                 No accessing organization means the request is already served in the tenant that owns the
+                 application, which is the case for the organization qualified and tenant qualified
+                 endpoints of the organization itself and for the root organization. The token was issued
+                 in that same tenant, so the key to validate it with is the application owner tenant key
+                 and the issuer claim does not need to be consulted.
+                */
                 oAuthAppDO = OAuth2Util.getAppInformationByClientId(clientId);
+                return OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO);
             }
             String appTenantDomain = OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO);
 
--- a/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
+++ b/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
@@ -529,13 +529,21 @@ public class OIDCLogoutServlet extends HttpServlet {
             String accessingOrgId = PrivilegedCarbonContext.getThreadLocalCarbonContext()
                     .getApplicationResidentOrganizationId();
             /*
-             The tenant qualified organization endpoints are served in the root tenant, where a lookup by
-             client id alone cannot find an application registered in a sub organization, so the application
-             is resolved through the organization hierarchy when an accessing organization is present.
+             The tenant qualified organization endpoints are served in the root tenant, so a lookup by client
+             id alone resolves against the root tenant and cannot find an application registered in a sub
+             organization. When an accessing organization is present the request is on one of those
+             endpoints, so the application is looked up in that organization's own tenant instead. Without
+             an accessing organization the request is already served in the tenant that owns the
+             application, and the existing lookup is used.
             */
-            OAuthAppDO oAuthAppDO = StringUtils.isNotEmpty(accessingOrgId)
-                    ? OAuth2Util.getAppInformationFromOrgHierarchy(clientId, accessingOrgId)
-                    : OAuth2Util.getAppInformationByClientId(clientId);
+            OAuthAppDO oAuthAppDO;
+            if (StringUtils.isNotEmpty(accessingOrgId)) {
+                String accessingTenantDomain = OIDCSessionManagementComponentServiceHolder.getInstance()
+                        .getOrganizationManager().resolveTenantDomain(accessingOrgId);
+                oAuthAppDO = OAuth2Util.getAppInformationByClientId(clientId, accessingTenantDomain);
+            } else {
+                oAuthAppDO = OAuth2Util.getAppInformationByClientId(clientId);
+            }
             String appTenantDomain = OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO);
 
             // The application owner tenant, which issues when the request is served in that tenant.
--- a/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
+++ b/components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
@@ -84,8 +84,6 @@ import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.UnsupportedEncodingException;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.security.interfaces.RSAPublicKey;
 import java.text.ParseException;
 import java.util.Enumeration;
@@ -97,8 +95,6 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.servlet.Servlet;
 import javax.servlet.ServletException;
@@ -129,10 +125,6 @@ import static org.wso2.carbon.identity.oidc.session.util.OIDCSessionManagementUt
 public class OIDCLogoutServlet extends HttpServlet {
 
     private static final Log log = LogFactory.getLog(OIDCLogoutServlet.class);
-    private static final Pattern TENANT_QUALIFIED_ISSUER_PATTERN =
-            Pattern.compile("/t/([^/]+)/oauth2/token$");
-    private static final Pattern ORGANIZATION_QUALIFIED_ISSUER_PATTERN =
-            Pattern.compile("/o/([^/]+)/oauth2/token$");
     private static final String REQUEST_PARAM_SP = "sp";
     private static final String UTF_8 = "UTF-8";
     private static final long serialVersionUID = -9203934217770142011L;
@@ -555,47 +547,11 @@ public class OIDCLogoutServlet extends HttpServlet {
             }
             String appTenantDomain = OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO);
 
-            // The tenant of the token issuer configured on the application.
-            IssuerDetails issuerDetails = oAuthAppDO.getIssuerDetails();
-            if (issuerDetails != null && StringUtils.isNotEmpty(issuerDetails.getIssuerTenantDomain())
-                    && tokenIssuer.equals(OAuth2OIDCConfigOrgUsageScopeUtils
-                            .getIssuerLocation(issuerDetails.getIssuerTenantDomain()))) {
-                return issuerDetails.getIssuerTenantDomain();
-            }
-            /*
-             An application that predates the token issuer setting has no issuer configured, which is the
-             state of applications migrated from before the setting existed. Issuance then falls back to the
-             tenant serving the request, so the root organization of this application's own organization is
-             also a legitimate issuer for it. That is the same pair of tenants the setting itself allows.
-            */
-            if (issuerDetails == null || StringUtils.isEmpty(issuerDetails.getIssuerTenantDomain())) {
-                OrganizationManager organizationManager = OIDCSessionManagementComponentServiceHolder
-                        .getInstance().getOrganizationManager();
-                String primaryTenantDomain = organizationManager.resolveTenantDomain(organizationManager
-                        .getPrimaryOrganizationId(organizationManager.resolveOrganizationId(appTenantDomain)));
-                if (tokenIssuer.equals(OAuth2OIDCConfigOrgUsageScopeUtils
-                        .getIssuerLocation(primaryTenantDomain))) {
-                    return primaryTenantDomain;
-                }
-            }
-            /*
-             The qualified issuer forms, /t/<tenant-domain>/oauth2/token and /o/<organization-id>/oauth2/token,
-             which are used when the relying party obtains the token through those endpoints. Both name a
-             tenant in the path. That tenant must be one of the two tenants that may issue for this
-             application, so that an issuer naming an unrelated tenant is not accepted, and the issuer must
-             have an origin this server builds.
-            */
-            String issuerTenant = tenantNamedByIssuer(tokenIssuer);
-            if (StringUtils.isNotEmpty(issuerTenant)
-                    && hasSameOrigin(tokenIssuer, OAuth2OIDCConfigOrgUsageScopeUtils
-                            .getIssuerLocation(appTenantDomain))) {
-                if (StringUtils.equals(appTenantDomain, issuerTenant)) {
-                    return appTenantDomain;
-                }
-                if (issuerDetails != null
-                        && StringUtils.equals(issuerDetails.getIssuerTenantDomain(), issuerTenant)) {
-                    return issuerTenant;
-                }
+            // The tenant that issues id tokens for this application. Its own issuer location is accepted
+            // as the issuer of the token.
+            String issuingTenantDomain = resolveIssuingTenantDomain(oAuthAppDO, appTenantDomain);
+            if (tokenIssuer.equals(OAuth2OIDCConfigOrgUsageScopeUtils.getIssuerLocation(issuingTenantDomain))) {
+                return issuingTenantDomain;
             }
             if (log.isDebugEnabled()) {
                 log.debug("The issuer of the id token does not match any issuer of the application. Client id: "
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
```diff
@@ -453,76 +454,112 @@ private String processLogoutRequest(HttpServletRequest request, HttpServletRespo
...
+                    }
+                    return true;
+                }
+                if (log.isDebugEnabled()) {
+                    log.debug("ID token signature did not match the tenant domain: " + tenantDomain);
+                }
+            } catch (JOSEException | ParseException e) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Error occurred while validating id token signature for the id token: " + idToken);
+                }
```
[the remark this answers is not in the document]
RushanNanayakkara: Valid, and partly introduced here: the previous commit moved this log inside the loop over candidate tenant domains, so the token could be written out more than once per request.

Fixed in b0e36d9b80. This message now records the tenant domain that was tried plus the exception, which is also more useful for diagnosis than the token. The two extraction failure messages in `getCandidateTenantDomainsForSignatureValidation` were changed the same way, so no complete id token is written anywhere in this path. Confirmed at runtime on a 7.3.0 pack: zero occurrences of a serialised token in the log after the change.

components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
```diff
@@ -477,52 +491,167 @@ private boolean validateIdToken(String idToken) {
...
+            }
+            String accessingOrgId = PrivilegedCarbonContext.getThreadLocalCarbonContext()
+                    .getApplicationResidentOrganizationId();
+            /*
+             The tenant qualified organization endpoints are served in the root tenant, where a lookup by
+             client id alone cannot find an application registered in a sub organization, so the application
+             is resolved through the organization hierarchy when an accessing organization is present.
+            */
+            OAuthAppDO oAuthAppDO = StringUtils.isNotEmpty(accessingOrgId)
+                    ? OAuth2Util.getAppInformationFromOrgHierarchy(clientId, accessingOrgId)
```
SujanSanjula96: We don't need to resolve this from hierarchy. Resolving from hierarchy needed only during authentication paths. We can use 
`OAuth2Util.getAppInformationByClientId(clientId, tenantDomain)` 
Here tenantDomain is the respective tenant domain of the accessing org ID

components/org.wso2.carbon.identity.oidc.session/src/main/java/org/wso2/carbon/identity/oidc/session/servlet/OIDCLogoutServlet.java
```diff
@@ -477,52 +491,167 @@ private boolean validateIdToken(String idToken) {
...
+            String clientId = extractClientFromIdToken(idToken);
+            String tokenIssuer = extractIssuerFromIdToken(idToken);
+            if (StringUtils.isBlank(tokenIssuer)) {
                 if (log.isDebugEnabled()) {
-                    log.debug("JWT signature will be validated with user tenant domain.");
+                    log.debug("The id token does not carry an issuer claim. Client id: " + clientId);
+                }
+                return null;
+            }
+            String accessingOrgId = PrivilegedCarbonContext.getThreadLocalCarbonContext()
```
SujanSanjula96, line 513: We can check like this,

If accessing org id not present,
fallback to old logic to get the app tenant domain using `OAuth2Util.getAppInformationByClientId(clientId)`
This will preserve the previous behaviours for `/o/<org-id>` and `/t/<tenant>` paths.

If accessing org id is  present, (sub org app logout request case with /t/<tenant>/o/<org-id> path)
get the corresponding tenant domain for accessing org ID and then get the oauth app using `OAuth2Util.getAppInformationByClientId(clientId, tenantDomain)` and get the tenant domain of app using the `OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO)`.
then from that we can check the issuer details and get the issuer tenant domain with a logic like
```

            IssuerDetails issuerDetails = oAuthAppDO.getIssuerDetails();
            if (issuerDetails != null && StringUtils.isNotEmpty(issuerDetails.getIssuerTenantDomain())
                    && tokenIssuer.equals(OAuth2OIDCConfigOrgUsageScopeUtils
                            .getIssuerLocation(issuerDetails.getIssuerTenantDomain()))) {
                return issuerDetails.getIssuerTenantDomain();
            }
```
(this is the flow we need to support issuer selection support. `/o/<org-id>` and `/t/<tenant>` paths are not related to that feature)

RushanNanayakkara: The structure you describe is already what the code does, so nothing changed for that part:

- no accessing organization id, fall back to `OAuth2Util.getAppInformationByClientId(clientId)`
- accessing organization id present, look up the app, then `OAuth2Util.getTenantDomainOfOauthApp(oAuthAppDO)`
- then the `issuerDetails` check, which is the same logic as the snippet you pasted

The lookup change is the same item as your other comment and is done in `addressing comments`.

On `/o/<org-id>` and `/t/<tenant>` not being related to issuer selection, I would keep those branches, because the measurements say the paths do interact with it:

- the endpoint used at issuance overrides the setting. With Token Issuer set to the root organization, a token obtained at `/o/<org-id>/oauth2/token` is signed with the **sub organization** key, not the root key.
- the relying party can log out on a different form from the one it obtained the token on. `/t/<org-handle>` mint then `/t/<root>/o/<org-id>` logout, and `/t/<root>/o/<org-id>` mint then `/o/<org-id>` logout, both fail on stock and are repaired by these branches.

An earlier revision of this PR had only the first two checks. It denied 7 cases that succeed on stock, including two controls that are plain unmodified tokens at their own endpoint, so it was a regression rather than a preserved behaviour.

Current state across four suites, all against stock: 7.3.0 functional 9 fixed 0 regressed, 7.3.0 cold start 5 fixed 0 regressed, 7.4.0-SNAPSHOT functional 5 fixed 0 regressed, 7.4.0-SNAPSHOT cold start 5 fixed 0 regressed. Happy to walk through the matrix if useful.

SujanSanjula96 (approved):

PR #3306, opened 2026-09-04
Fix object casting issue

NutharaNR:
#### Root Cause

The failure occurs at [1], where the `realm` claim is cast directly to a concrete `HashMap`:

```java
Map<String, String> realm = (HashMap) claimsSet.getClaim(OAuthConstants.OIDCClaims.REALM);
```

With Nimbus 10, an object-valued claim may be represented as a `LinkedTreeMap` rather than a `HashMap`. As a result, the concrete cast can throw a `ClassCastException`.

Additionally, the `realm` claim is not always an object. For example, it may be a string such as `"/alpha"`. Attempting to cast such a value to a map will also result in an exception.

#### Fix

Cast the claim to the `Map` interface and verify its type using `instanceof` before processing it. This follows the same pattern already used in `OIDCLogoutServlet.extractTenantDomainFromIdToken`:

```java
Map realm = null;

if (claimsSet.getClaim(OAuthConstants.OIDCClaims.REALM) instanceof Map) {
    realm = (Map) claimsSet.getClaim(OAuthConstants.OIDCClaims.REALM);
}
```

This approach handles both scenarios:

* **Map-valued realm claims:** The `Map` interface supports implementations such as Nimbus 10's `LinkedTreeMap`.
* **Non-object realm claims:** Values such as `"/alpha"` are ignored instead of causing a `ClassCastException`. This comes as a use case where we set the "realm" value via Pre Issue Access Token Action.

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/blob/9a7817571ce29646ce3e0e2bdaeaea4c287489c4/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/JWTUtils.java#L435

### DIFF
--- a/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/JWTUtils.java
+++ b/components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/JWTUtils.java
@@ -61,7 +61,6 @@ import java.security.cert.X509Certificate;
 import java.security.interfaces.RSAPublicKey;
 import java.text.ParseException;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -432,18 +431,21 @@ public class JWTUtils {
     public static String getSigningTenantDomain(JWTClaimsSet claimsSet, AccessTokenDO accessTokenDO)
             throws ParseException, IdentityOAuth2Exception {
 
-        Map<String, String> realm = (HashMap) claimsSet.getClaim(OAuthConstants.OIDCClaims.REALM);
+        Map realm = getRealmClaim(claimsSet);
         if (MapUtils.isNotEmpty(realm)) {
-            if (realm.get(OAuthConstants.OIDCClaims.SIGNING_TENANT) != null) {
+            String signingTenant = getClaimWithinRealmAsString(realm, OAuthConstants.OIDCClaims.SIGNING_TENANT);
+            if (signingTenant != null) {
                 if (log.isDebugEnabled()) {
                     log.debug("Getting signing tenant domain from JWT's 'signing_tenant' claim.");
                 }
-                return realm.get(OAuthConstants.OIDCClaims.SIGNING_TENANT);
-            } else if (realm.get(OAuthConstants.OIDCClaims.TENANT) != null) {
+                return signingTenant;
+            }
+            String tenant = getClaimWithinRealmAsString(realm, OAuthConstants.OIDCClaims.TENANT);
+            if (tenant != null) {
                 if (log.isDebugEnabled()) {
                     log.debug("Getting signing tenant domain from JWT's 'tenant' claim.");
                 }
-                return realm.get(OAuthConstants.OIDCClaims.TENANT);
+                return tenant;
             }
         }
         if (accessTokenDO == null) {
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/JWTUtils.java
```diff
@@ -432,18 +431,22 @@ public static int getSubOrgStartLevel() {
...
                     log.debug("Getting signing tenant domain from JWT's 'signing_tenant' claim.");
                 }
-                return realm.get(OAuthConstants.OIDCClaims.SIGNING_TENANT);
+                return (String) realm.get(OAuthConstants.OIDCClaims.SIGNING_TENANT);
             } else if (realm.get(OAuthConstants.OIDCClaims.TENANT) != null) {
                 if (log.isDebugEnabled()) {
                     log.debug("Getting signing tenant domain from JWT's 'tenant' claim.");
                 }
-                return realm.get(OAuthConstants.OIDCClaims.TENANT);
+                return (String) realm.get(OAuthConstants.OIDCClaims.TENANT);
```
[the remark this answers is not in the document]
NutharaNR: fixed

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/JWTUtils.java
```diff
@@ -432,18 +431,25 @@ public static int getSubOrgStartLevel() {
     public static String getSigningTenantDomain(JWTClaimsSet claimsSet, AccessTokenDO accessTokenDO)
             throws ParseException, IdentityOAuth2Exception {
 
-        Map<String, String> realm = (HashMap) claimsSet.getClaim(OAuthConstants.OIDCClaims.REALM);
+        Map realm = null;
+        if (claimsSet.getClaim(OAuthConstants.OIDCClaims.REALM) instanceof Map) {
+            //
```
piraveena: remove this //

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/JWTUtils.java
```diff
@@ -499,6 +505,21 @@ public static String getSigningTenantDomain(JWTClaimsSet claimsSet, AccessTokenD
...
+    /**
+     * Reads a nested value out of the 'realm' claim, returning it only when it is a String. A 'realm' entry with
+     * any other JSON type is treated as absent so that a malformed token falls back to the other resolution paths
+     * instead of failing the whole validation flow.
+     *
+     * @param realm     The 'realm' claim of the JWT.
+     * @param claimName The nested claim to read.
+     * @return The claim value, or null if it is missing or not a String.
+     */
+    private static String getRealmStringValue(Map realm, String claimName) {
```
piraveena: getStringClaimFromRealm or something to give clear meaning?

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth2/util/JWTUtils.java
piraveena: Shall we have a separate private method to get the realm claim? It can be used here and in the other methd too

components/org.wso2.carbon.identity.oauth/.java-version
```diff
@@ -0,0 +1 @@
+21
```
piraveena: remove this

piraveena (approved):

PR #3308, opened 2026-09-11
 Add application level scope restriction property for token exchange federated tokens

RovinKYK:
## Purpose

Add the application-level configuration that restricts scope issuance for federated subject tokens in the token exchange grant, and seed its value for new applications from the tenant default.

Related Issue: https://github.com/wso2/product-is/issues/28406

## Description

Adds `tokenExchangeRestrictScopeIssuanceForFederatedTokens` as an OIDC SP property. No schema change, it is a row in `IDN_OIDC_PROPERTY`.

- `OAuthAppDO` and `OAuthConsumerAppDTO` carry the value as a nullable `Boolean`, so an omitted request field is distinguishable from an explicit `false`.
- `OAuthAppDAO` writes it on create and update only when non-null, and leaves the field untouched when the property row is absent. Applications that predate this change therefore read `null` and keep today's behaviour.
- `OAuthAdminServiceImpl` stamps the value during registration: an explicit value in the request always wins, and only when none is given is the tenant's `defaultRestrictScopeIssuanceForFederatedTokens` compatibility setting consulted. 
- Adds a `CompatibilitySettingsService` OSGi reference and the corresponding service holder accessors.

Unit tests: DAO round-trip for both values, DAO read when the property row is absent, and tenant-default resolution across setting values.


components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/internal/OAuthServiceComponent.java
```diff
@@ -614,4 +615,23 @@ protected void unregisterActionExecutorService(ActionExecutorService actionExecu
...
+    protected void setCompatibilitySettingsService(CompatibilitySettingsService compatibilitySettingsService) {
+
+        log.debug("Setting the CompatibilitySettingsService in OAuthServiceComponent.");
+        OAuthComponentServiceHolder.getInstance().setCompatibilitySettingsService(compatibilitySettingsService);
+    }
+
+    protected void unsetCompatibilitySettingsService(CompatibilitySettingsService compatibilitySettingsService) {
+
+        log.debug("Unsetting the CompatibilitySettingsService in OAuthServiceComponent.");
+        OAuthComponentServiceHolder.getInstance().setCompatibilitySettingsService(null);
```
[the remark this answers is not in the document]
RovinKYK: Every singleton unbind in this class nulls the holder, and since the reference is MANDATORY the component isn't active without the service.

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
```diff
@@ -3700,4 +3714,38 @@ private void normalizeGracefulReuseLimit(OAuthAppDO appDO) {
...
+                    compatibilitySettingsService.getCompatibilitySettingsByGroupAndSetting(tenantDomain,
+                            TOKEN_EXCHANGE_COMPATIBILITY_SETTING_GROUP,
+                            DEFAULT_RESTRICT_FEDERATED_TOKEN_SCOPE_ISSUANCE_COMPATIBILITY_KEY);
+            CompatibilitySettingGroup group =
+                    setting.getCompatibilitySetting(TOKEN_EXCHANGE_COMPATIBILITY_SETTING_GROUP);
+            if (group == null) {
+                return false;
+            }
+            return Boolean.parseBoolean(
+                    group.getSettingValue(DEFAULT_RESTRICT_FEDERATED_TOKEN_SCOPE_ISSUANCE_COMPATIBILITY_KEY));
```
hwupathum: We can combine this logic

components/org.wso2.carbon.identity.oauth/src/main/java/org/wso2/carbon/identity/oauth/OAuthAdminServiceImpl.java
```diff
@@ -3700,4 +3714,38 @@ private void normalizeGracefulReuseLimit(OAuthAppDO appDO) {
...
+     * Check whether restricting scopes for federated subject tokens is enabled by default for the given tenant.
+     *
+     * @param tenantDomain Tenant domain of the application.
+     * @return true if the tenant default is enabled, false otherwise.
+     */
+    private boolean isRestrictScopeIssuanceEnabledByDefault(String tenantDomain) {
+
+        CompatibilitySettingsService compatibilitySettingsService =
+                OAuthComponentServiceHolder.getInstance().getCompatibilitySettingsService();
+        if (compatibilitySettingsService == null || StringUtils.isBlank(tenantDomain)) {
```
hwupathum: We can check for tenantDomain null at the beginning

hwupathum (approved):

